PR #2, opened 2020-02-13
Matilda Arvidsson - happy thoughts

HIPPIEKICK:
Fun week! Wish I had time to try out more stretch goals, and really think about what I've done. I will try to have a look at this again when I have more time. Wondering if I need more error messages / response statuses.

### DIFF
--- a/server.js
+++ b/server.js
@@ -11,16 +11,39 @@ mongoose.Promise = Promise
 // overridden when starting the server. For example:
 //
 //   PORT=9000 npm start
-const port = process.env.PORT || 8080
+const port = process.env.PORT || 8000
 const app = express()
 
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
+//Model
+const Thought = mongoose.model("Thought", {
+  message: {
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -11,16 +11,76 @@ mongoose.Promise = Promise
...
+    res.status(201).json(savedThought)
+  }
+  catch (err) {
+    res.status(400).json({ message: "Couldn't save thought", error: err.errors })
+  }
+})
+
+app.post("/:thoughtID/like", async (req, res) => {
+  //Find specific thought
+  const thoughtID = req.params.thoughtID
```
JennieDalgren: `const { thoughtID } = req.params` you can be consistent and do it this way here as well

server.js
```diff
@@ -11,16 +11,76 @@ mongoose.Promise = Promise
...
+  hearts: {
+    type: Number,
+    default: 0
+  },
+  name: {
+    type: String,
+    default: "anonymous",
+    minlength: 2,
+    maxlength: 50
+  },
```
JennieDalgren: Nice little add here!

PR #11, opened 2020-02-16
Happy-Thoughts-API

Nasimmhn:
Link to my deployed project:
Backend:  https://happy-thoughts-api-by-nasim.herokuapp.com/
Frontend: https://happy-thoughts-by-nasim.netlify.com/

Things I'd like to have clarified or explained in more detail:
- Is it necessary to write  (mongoose.Promise = Promise)? if yes, why? I just didn't use it, but it still works. 
Thanks/
Nasim

### DIFF
--- a/server.js
+++ b/server.js
@@ -6,7 +6,7 @@ import mongoose from 'mongoose'
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
 
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
- 
+
 
 const Thought = mongoose.model('Thought', {
   message: {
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -23,6 +40,54 @@ app.get('/', (req, res) => {
...
+
+  try {
+    //Success
+    const like = await Thought.findOneAndUpdate(
+      { "_id": req.params.thoughtId },//filter
+      { $inc: { "heart": 1 } },//update
+      { returnNewDocument: true }//doesn't update/work
+
+    )
+    res.status(201).json(like)
```
puj: I like that you returned the new thought here!  Hopefully this worked as expected!
This will allow the frontend to rely on the backend for the updated thought information

server.js
```diff
@@ -4,12 +4,29 @@ import cors from 'cors'
...
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
-mongoose.Promise = Promise
+ 
+
+const Thought = mongoose.model('Thought', {
+  message: {
+    type: String,
+    required: true,
+    minlength: 5,
+    maxlength: 140
```
puj: Matches the frontent, that's good.  I wonder how you could tell the frontend about these requirements instead of "hardcoding" them in two different places 🤔

PR #13, opened 2020-02-16
project-happy-thougths-api - Susanna Sjöberg

SanSjo:
https://happy-thoughts-app.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -36,18 +36,15 @@ app.use(cors());
 app.use(bodyParser.json());
 
 // Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world');
-});
 
-app.get('/thoughts', async (req, res) => {
+app.get('/', async (req, res) => {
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -1,29 +1,77 @@
...
+    });
+  }
+});
+
+app.post('/:id/like', async (req, res) => {
+  try {
+    const thoughtLiked = await Thought.findOneAndUpdate(
+      { _id: req.params.id },
+      { $inc: { heart: 1 } }
+    );
```
puj: Does this return the new or the old thought?

server.js
```diff
@@ -1,29 +1,77 @@
...
+    res.status(201).json(thoughtLiked);
+  } catch (err) {
+    res.status(404).json({ message: 'could not add like', error: err.errors });
```
puj: 404 is good here

PR #14, opened 2020-02-16
Happy thpughts API - Anna Almestål

anna-ip:
Had some problems this week, first making the like endpoint to work, realised that I've searched for the wrong solution, but thanks to great class mates I managed to get on the right path again 😬
Then I've had some struggles connecting to Heruko, but now it all should work, both frontend and back end.
Think its very interesting working with the backend, easier to see the connections with the frontend as well.
https://happy-thought-api.herokuapp.com/
https://happy-thought-messages-anna-almestal.netlify.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -6,6 +6,7 @@ import mongoose from 'mongoose'
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
+mongoose.set('useFindAndModify', false);
 
 const Thought = mongoose.model('Thought', {
   message: {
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -19,8 +53,40 @@ app.use(cors())
...
 // Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+// app.get('/', (req, res) => {
+//   res.send('Hello world')
+// })
+
+app.get('/', async (req, res) => {
+  const thought = await Thought.find().sort({ createdAt: 'desc' }).limit(20).exec()
+  res.json(thought)
```
puj: Is this a single thought or multiple?

server.js
```diff
@@ -19,8 +53,40 @@ app.use(cors())
...
+})
+
+app.post('/', async (req, res) => {
```
puj: It's not clear what a POST to '/' will do.  

Either naming the endpoint '/thoughts' or adding a comment to the code would be appropriate here.

PR #21, opened 2020-02-17
Happy API -Jenny Jacobsson

jennyjacobsson:
Hi,

This was a fun project, it was nice to work with both front- and backend, and then try to put it all  together. :) 

Frontend: https://jenny-project-11-happythoughts.netlify.com/

Backend: https://jenny-happy-api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -7,6 +7,25 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
+const Thought = mongoose.model('Thought', {
+  message: {
+    type: String,
+    required: true,
+    minlength: 5,
+    maxlength: 140
+  },
+  hearts: {
+    type: Number,
+    default: 0
+    ///Should not be assignable when creating a new thought
+  },
+  createdAt: {
+    type: Date,
+    default: Date.now
+    //Should not be assignable when creating a new thought
+  },
+})
+
 // Defines the port the app will run on. Defaults to 8080, but can be 
 // overridden when starting the server. For example:
 //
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -7,6 +7,32 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
 mongoose.Promise = Promise
 
+const Thought = mongoose.model('Thought', {
+
+  name: {
+    type: String,
+    maxlength: 50,
+    default: 'Anonymous'
+  },
+
```
puj: Cool!

server.js
```diff
@@ -18,11 +44,37 @@ const app = express()
...
+  } else {
+    res.status(400).json({ message: "Fill up this empty space with your happiness" })
+  }
+})
+
+app.post('/:thoughtId/like', async (req, res) => {
+  try {
+    const { thoughtId } = req.params
+    await Thought.updateOne({ '_id': thoughtId }, { '$inc': { 'hearts': 1 } })
+    res.status(201).json({ message: `Thought Id nr: ${thoughtId} got one heart update` })
```
puj: Why not return the updated Thought?

server.js
```diff
@@ -18,11 +44,37 @@ const app = express()
...
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+
+app.get('/', async (req, res) => {
+  const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20).exec()
+  if (thoughts.length > 0) {
+    res.json(thoughts)
+  } else {
+    res.status(400).json({ message: "Fill up this empty space with your happiness" })
```
puj: Do you really want to return a 400 if the user just started using your site for the first time?

server.js
```diff
@@ -18,11 +44,37 @@ const app = express()
...
+  }
 })
 
+app.post('/', async (req, res) => {
+  try {
+    const { name, message } = req.body
+    const thought = await new Thought({ message, name }).save()
+    res.status(201).json(thought)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not save thought', errors: err.error })
```
puj: Is that because it can't be found?  Maybe 404 is better?

PR #25, opened 2020-02-17
Happy-twitterish-api

johannawaern:
I think backend is kind of fun, it's nice to be able to see how things are connected. However, I guess I wont be working as a backend developer in the nearest future or ever, but I like the fact that I have a tiny bit of knowledge and I look forward to the final project when we have to use this experiences in a bigger project. 

(just realised that my errors does not look good in the frontend :), looking forward to have some time over and go back to this app and do both frontend and backend properly!)

### DIFF
--- a/server.js
+++ b/server.js
@@ -44,6 +44,19 @@ app.get('/thoughts', async (req, res) => {
   res.json(thoughts)
 })
 
+app.post('/thoughts', async (req, res) => {
+  const { message, heart } = req.body
+  const thought = new Thought({ message, heart })
+  try {
+    const savedThought = await thought.save()
+    res.status(201).json(savedThought)
+  } catch (err) {
+    res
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -23,6 +36,38 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+app.get('/thoughts', async (req, res) => {
+  const thoughts = await Thought.find()
+    .sort({ createdAt: 'desc' })
+    .limit(20)
+    .exec()
```

server.js
```diff
@@ -23,6 +36,38 @@ app.get('/', (req, res) => {
...
+  res.json(thoughts)
+})
+
+app.post('/thoughts', async (req, res) => {
+  const { message } = req.body
+  const thought = new Thought({ message })
```

PR #29, opened 2020-02-19
My happy thoughts API

JoacimNilsen:
Fun project implementing backend with our previous front end!! I had some troubles with the database again and think it is a bit abstract how it works. Got it to work tho! I have an issue with the likes not being saved to DB and would like that clarified. I think it is in the front-end code and will continue to find the issue. Could use a pointer or two :)

https://joacims-happy-thoughts.herokuapp.com/

front-end: https://serene-blackwell-83b86d.netlify.com/

### DIFF
--- a/models/thought.js
+++ b/models/thought.js
@@ -1,0 +1,18 @@
+import mongoose from 'mongoose'
+
+export const Thought = mongoose.model('Thought', {
+  message: {
+    type: String,
+    required: true,
+    minlength: 2,
+    maxlength: 140
+  },
+  likes: {
+    type: Number,
+    default: 0
+  },
+  createdAt: {
+    type: Date,
+    default: Date.now
+  }
+})
--- a/models/thought.js
+++ b/models/thought.js
@@ -4,7 +4,7 @@ export const Thought = mongoose.model('Thought', {
   message: {
     type: String,
     required: true,
-    minlength: 2,
+    minlength: 5,
     maxlength: 140
   },
   likes: {
--- a/server.js
+++ b/server.js
@@ -3,14 +3,12 @@ import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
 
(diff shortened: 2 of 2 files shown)


models/thought.js
```diff
@@ -0,0 +1,18 @@
+import mongoose from 'mongoose'
```
JennieDalgren: Nice that you've extracted this out to its own file 👍

server.js
```diff
@@ -3,24 +3,49 @@ import bodyParser from 'body-parser'
...
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+// My routes
+app.get('/', async (req, res) => {
+  const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20).exec()
```

server.js
```diff
@@ -3,24 +3,49 @@ import bodyParser from 'body-parser'
...
+// POST thoughts to page
+app.post('/', async (req, res) => {
+    const { message } = req.body
+    try {
+    const thought = await new Thought({ message }).save()
+    res.status(200).json(thought)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not post thought', error: err.error })
+  }
+})
```
JennieDalgren: messy indentation here 😉

server.js
```diff
@@ -3,24 +3,49 @@ import bodyParser from 'body-parser'
...
+
+// POST likes to page
+app.post('/:thoughtId/likes', async (req, res) => {
+  const { thoughtId } = req.params
+  try {
+    await Thought.updateOne({'_id': thoughtId }, {$inc: {'likes': 1}})
```
JennieDalgren: The only thing might be that you should remove the ' ' around likes?
`{ $inc: { likes: 1 } }`
other than that, this endpoint seems to be correct.

PR #31, opened 2020-02-24
Late PR

ucha3i:
It seems to work with frontend, I only have to changed the link there. I need to test it better, though. 

backend: https://happythoughtsapi.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -7,10 +7,23 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
-// overridden when starting the server. For example:
-//
-//   PORT=9000 npm start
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -7,22 +7,72 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
-app.get('/', (req, res) => {
+// Start defining your routes here -> code along with Jennie
+app.get('/', (_, res) => {
   res.send('Hello world')
 })
 
+app.get('/', async (_, res) => {
+  const thoughts = await Thought.find()
+    .sort({ createdAt: 'desc' })
+    .limit(20)
```

server.js
```diff
@@ -7,22 +7,72 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+  res.json(thoughts)
+})
+
+app.post('/', async (req, res) => {
+  const thought = new Thought({
+    message: req.body.message,
+    hearts: 0
```
JennieDalgren: you don't need to specify hearts: 0 here since your getting that as a default from the mongoose model

PR #39, opened 2020-05-18
Project Happy thoughts API - Christina Persson: Review wanted 🤗

Cilpersson:
# View it live: 
https://perssons-happy-thoughts.herokuapp.com/
## Frontend
https://perssons-happy-thoughts.netlify.app/

## My thoughts
 I loved The happy thoughts project and I loved going back to it even more. I might have gone a bit crazy with features but I have learned so much mongo stuff thanks to it so I'm keeping all the features! The only thing I'm not completely certain about is if I have to use the Schema-part when putting the model in it's own file and folder.

### DIFF
--- a/server.js
+++ b/server.js
@@ -55,7 +55,7 @@ app.get("/thoughts", async (req, res) => {
     .limit(perPage)
     .skip(skip)
     .exec();
-  res.json([{ pages: pages }, thoughts]);
+  res.json([pages, thoughts]);
 });
 
 app.post("/thoughts", async (req, res) => {
--- a/server.js
+++ b/server.js
@@ -62,6 +62,21 @@ app.post("/thoughts", async (req, res) => {
   }
 });
 
+app.post("/:thoughtID/like", async (req, res) => {
+  const { thoughtID } = req.params;
+  const thought = await Thought.findById(thoughtID);
+
+  if (thought) {
+    thought.hearts++;
+    thought.save();
+    res.status(201).json(thought);
+  } else {
+    res.status(400).json({
+      message: `Could not find thought with ID: ${thoughtID}`,
+    });
+  }
+});
+
 // Start the server
 app.listen(port, () => {
   console.log(`Server running on http://localhost:${port}`);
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -1,29 +1,103 @@
...
+
+app.use(cors());
+app.use(bodyParser.json());
+
+const listEndpoints = require("express-list-endpoints");
+
+//Tried to use consts instead of text in error handling.
+const COULD_NOT_SAVE_THOUGHT = "Could not save thought to DB";
+const COULD_NOT_FIND_THOUGHT_WITH_ID = "Could not find thought with ID:";
+const COULD_FIND_THOUGHT = "Likes increased by one";
```
HIPPIEKICK, line 22: No biggie, but this variable could have a more descriptive name :)

HIPPIEKICK:
I approved this, but I noticed a little late that the endpoints should look like below. Please take a look when you have time, and update it :)

### `GET /`

This endpoint should return a maximum of 20 thoughts, sorted by `createdAt` to show the most recent thoughts first.

### `POST /`

This endpoint expects a JSON body with the thought `message`, like this: `{ "message": "Express is great!" }` . If the input is valid (more on that below), the thought should be saved, and the response should include the saved thought object, including its `_id`.

### `POST /:thoughtId/like`

Cilpersson:
> I approved this, but I noticed a little late that the endpoints should look like below. Please take a look when you have time, and update it :)
> 
> ### `GET /`
> This endpoint should return a maximum of 20 thoughts, sorted by `createdAt` to show the most recent thoughts first.
> 
> ### `POST /`
> This endpoint expects a JSON body with the thought `message`, like this: `{ "message": "Express is great!" }` . If the input is valid (more on that below), the thought should be saved, and the response should include the saved thought object, including its `_id`.
> 
> ### `POST /:thoughtId/like`
> This endpoint doesn't require a JSON body. Given a valid thought id in the url, the API should find that thought, and update its `hearts` property to add one heart.

I don't really get what in my endpoints that is incorrect 🙈I have done pagination so there are 20 thoughts per page, is it that I have POST & GET /thoughts rather than just POST & GET / ?

PR #40, opened 2020-05-19
Project week 19, Ingrid H

ingrid2019-ism:


server.js
```diff
@@ -18,9 +33,36 @@ const app = express()
 app.use(cors())
 app.use(bodyParser.json())
 
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+// Get the 20 most recent thoughts
+app.get('/', async (req, res) => {
+  const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20).exec()
```
HIPPIEKICK, line 38: ⭐

server.js
```diff
@@ -18,9 +33,36 @@ const app = express()
...
+  res.json(thoughts)
+})
+
+// Send a new thought
+app.post('/', async (req, res) => {
+  const { message } = req.body
+  const thought = new Thought({
+    message: message,
+  })
```
HIPPIEKICK, line 47: When the key and value are the same, you could write like this:
`const thought = new Thought({ message })`

server.js
```diff
@@ -18,9 +33,36 @@ const app = express()
...
+  try {
+    const savedThought = await thought.save()
+    res.status(201).json(savedThought)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not save happy thought to database', error: err.errors })
```
HIPPIEKICK, line 52: Nice and clear error message!

server.js
```diff
@@ -18,9 +33,36 @@ const app = express()
...
+
+// Like a thought
+app.post('/:id/like', async (req, res) => {
+  try {
+    const savedLike = await Thought.findOneAndUpdate(
+      { _id: req.params.id }, { $inc: { hearts: 1 } }
+    )
+    res.json(savedLike)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not like, thought not found', error: err.errors })
```
HIPPIEKICK, line 64: Descriptive and clear!

server.js
```diff
@@ -18,9 +33,36 @@ const app = express()
...
+  try {
+    const savedThought = await thought.save()
+    res.status(201).json(savedThought)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not save happy thought to database', error: err.errors })
+  }
+})
+
+// Like a thought
+app.post('/:id/like', async (req, res) => {
```
HIPPIEKICK, line 57: Update this endpoint so it corresponds with the documentation:

### `POST /:thoughtId/like`

PR #42, opened 2020-05-19
review wanted

kajsabootin:
Link to my deployed project:

https://happy-thoughts-api-kajsa.herokuapp.com/

frondend: https://kajsa-happythought.netlify.app/

Things I'd like to have clarified or explained in more detail:

I had really problem to understand how I was able to check if "like" worked.

I am most proud of that this is my first project that is both backend and frontend. So crazy??????

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,29 +1,52 @@
-import express from 'express'
-import bodyParser from 'body-parser'
-import cors from 'cors'
-import mongoose from 'mongoose'
+import express from "express";
+import bodyParser from "body-parser";
+import cors from "cors";
+import mongoose from "mongoose";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
-mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
-mongoose.Promise = Promise
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts";
+mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
+mongoose.Promise = Promise;
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
+const Thought = mongoose.model("Thought", {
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,29 +1,87 @@
...
+});
+
+//Get the 20 latest thoughts route
+app.get("/thoughts", async (req, res) => {
+  const thoughts = await Thought.find()
+    .sort({ createdAt: "desc" })
+    .limit(20)
+    .exec();
+  res.json(thoughts);
+});
```
KarolCislowski, line 51: according to documentation` get "/" ` need to return 20 thoughts,

server.js
```diff
@@ -1,29 +1,87 @@
...
+  try {
+    // Succes
+    const savedThought = await thought.save();
+    res.status(201).json(savedThought);
+  } catch (err) {
+    res
+      .status(400)
+      .json({ message: "Sorry, could not post this", error: err.errors });
+  }
+});
```
KarolCislowski, line 70: according to documentation `post "/" ` need to create a new one.

server.js
```diff
@@ -1,29 +1,87 @@
...
+  try {
+    const thought = await Thought.updateOne(
+      { _id: req.params.thoughtId },
+      { $inc: { hearts: 1 } }
+    );
+    res.status(201).json(thought);
+  } catch (err) {
+    res.status(401).json({ message: "Heart not added to post", error: err });
+  }
+});
```
KarolCislowski, line 82: This one should by ` /:thoughtId/like `

server.js
kajsabootin: We got information from Matilda that "if you have received changes requested due to the names of the endpoints, we will change them to approved" . And if I understand right I haven't been approved regarding my endpoints name.

kajsabootin:
We got information from Matilda that "if you have received changes requested due to the names of the endpoints, we will change them to approved" . And if I understand right I haven't been approved regarding my endpoints name.

PR #43, opened 2020-05-20
Happy api

elinaarenstrup:


server.js
```diff
@@ -1,29 +1,95 @@
...
-app.use(bodyParser.json())
+app.use(cors());
+app.use(bodyParser.json());
 
 // Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+app.get("/", (req, res) => {
+  res.send(listEndpoints(app));
```
HIPPIEKICK, line 43: ⭐

server.js
```diff
@@ -1,29 +1,95 @@
...
+});
+
+//Get the 20 latest thoughts route, desc = newest first
+app.get("/thoughts", async (req, res) => {
+  const thoughts = await Thought.find()
+    .sort({ createdAt: "desc" })
+    .limit(20)
+    .exec();
+  res.json(thoughts);
```
HIPPIEKICK, line 52: Nice and clear!

server.js
```diff
@@ -1,29 +1,95 @@
...
+app.post("/thoughts", async (req, res) => {
+  // Retrive the information sent by the client to our API endpoint
+  const { message } = req.body;
+  // Then use our mongoose model to create the database entry
+  const thought = new Thought({ message });
+
+  try {
+    // Success-case, send good status code to the client
+    const savedThought = await thought.save({ message });
+    res.status(200).json(savedThought);
```
HIPPIEKICK, line 65: Since you put the `message` in the `thought` on line 60, you don't need it on row 64 as well:
```
const thought = new Thought({ message });
  try {
    const savedThought = await thought.save();
    res.status(200).json(savedThought);
  }
```

server.js
```diff
@@ -1,29 +1,95 @@
...
+  } catch (err) {
+    // Bad-req, send bad status code to the client
+    res.status(404).json({
```
HIPPIEKICK, line 68: Status 404 means "not found", 400 is "bad request" :)

server.js
```diff
@@ -1,29 +1,95 @@
...
+      message: "Sorry, could not save this thought to database.",
+      errors: err.errors,
+    });
+  }
+});
+
+// Post likes to a specific message
+// Using POST and not PUT because I increment likes, not adding new data to database.
+app.post("/:id/like", async (req, res) => {
```
HIPPIEKICK, line 77: Since your other endpoints are named `/thoughts` - you could name this `/thoughts/:id/like` to be more consistent

server.js
```diff
@@ -1,29 +1,95 @@
...
+  try {
+    const thought = await Thought.findOneAndUpdate(
+      { _id: req.params.id },
+      //increment by 1
+      { $inc: { hearts: 1 } }
+    );
```
HIPPIEKICK, line 83: 🥳

server.js
```diff
@@ -1,29 +1,95 @@
...
+    res.json(thought).status(200);
```
HIPPIEKICK, line 84: Why did you change from 201? :)

HIPPIEKICK:
I approved this, but I noticed a little late that the endpoints should look like below. Please take a look when you have time, and update it :)

### `GET /`

This endpoint should return a maximum of 20 thoughts, sorted by `createdAt` to show the most recent thoughts first.

### `POST /`

This endpoint expects a JSON body with the thought `message`, like this: `{ "message": "Express is great!" }` . If the input is valid (more on that below), the thought should be saved, and the response should include the saved thought object, including its `_id`.

### `POST /:thoughtId/like`

server.js
elinaarenstrup: I got super confused when I saw that 200 was "ok" and 201 was "created", so I went with 200. Which one of these is the correct one to use? "Ok" sounded more correct to me but not sure.

HIPPIEKICK: Exactly, 201 is "created" (and it's the most common status code to use for when POST-requests are successful). I think 200 is good to use when you're unsure, since it's broader than e.g. 201, so neither are incorrect in this case! I was just curious of why you changed it :)

PR #44, opened 2020-05-20
Happy Thoughts API - Emma

emmapre:


server.js
```diff
@@ -7,20 +7,91 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+  const totalThoughts = await Thought.find()
+  const pages = Math.ceil(totalThoughts.length / perPage)
+
+  const thoughts = await Thought.find()
+    .sort({ createdAt: 'desc' })
+    .limit(perPage)
+    .skip(skip)
+
+  res.json({ thoughts, pages })
+})
```
KarolCislowski, line 69: according to documentation `get "/"` need to return 20 thoughts,

server.js
```diff
@@ -7,20 +7,91 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+app.post('/thoughts', async (req, res) => {
+  const { message, name } = req.body
+  const thought = new Thought({ message, name })
+  try {
+    const saveThought = await thought.save()
+    res.status(201).json(saveThought)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not post thought', error: err.errors })
+  }
+})
```
KarolCislowski, line 82: according to documentation `post "/"` need to create a new one.

server.js
```diff
@@ -7,20 +7,91 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+// return error if thought not found
+//CHANGE TO PUT
+app.post('/:thoughtId/like', async (req, res) => {
+  const { thoughtId } = req.params
+  try {
+    await Thought.updateOne({ _id: thoughtId }, { $inc: { hearts: 1 } })
+    res.status(201).json()
+  } catch (err) {
+    res.status(404).json({ message: 'Thought not found', error: err.errors })
+  }
```
KarolCislowski, line 94: This one should by /:thoughtId/like

server.js
emmapre: It's paginated so it should only return 20 pages. Is it not working?

server.js
emmapre: I don't understand, the endpoint is /:thoughtId/like?

PR #47, opened 2020-05-20
project happy thoughts api - Jennifer

Jenniferot:
It was fun to learn how to connect frontend with backend. I struggled a bit, but finally understood most of it. I made one stretch goal and added a name field to each thought. I still have some green code left since I started adding a tag as well but hadn't enough time to finish that.

Backend: https://happylove-api.herokuapp.com
Frontend: https://inspiring-goodall-3fb1f9.netlify.app


server.js
```diff
@@ -18,11 +18,78 @@ const app = express()
...
+  createdAt: {
+    type: Date,
+    default: Date.now
+  },
+  tag: {
+    type: String
+  },
+  name: {
+    type: String,
+  }
```
KarolCislowski, line 42: I think you should make this field required ;)

server.js
```diff
@@ -18,11 +18,78 @@ const app = express()
...
+  message: {
+    type: String,
+    required: true,
+    minlength: 5,
+    maxlength: 120
+  },
+  hearts: {
+    type: Number,
+    required: true,
+  },
```
KarolCislowski, line 32: Here you could use `default : 0` instead of `required`

server.js
```diff
@@ -18,11 +18,78 @@ const app = express()
...
+    .sort({ createdAt: "desc" })
+    .limit(20)
+  const { tag } = req.query
+  if (tag) {
+    const filteredThoughts = await Thought.find({ tag: tag })
+    res.json(filteredThoughts)
+  } else {
+    res.json(thoughts)
+    //res.send('Hello world')
+  }
```
KarolCislowski, line 57: Some error handling would be nice :)

server.js
```diff
@@ -18,11 +18,78 @@ const app = express()
...
+      .json({ message: 'could not save thought', errors: err.errors })
+  }
+})
+
+app.post('/:id/like',
+  async (req, res) => {
+    try {
+      const thought = await Thought.findOneAndUpdate(
+        { _id: req.params.id },
+        { $inc: { hearts: 1 } },
```
KarolCislowski, line 82: Nice incrementation :)

PR #50, opened 2020-05-20
Project 19 - Johanna Åstrand

jastrand:
This is my project! Getting the likes to work was a bit tricky and I would have loved to spend some time to add more features but here it is: 

Frontend: https://infallible-archimedes-9a857b.netlify.app/
Backend: https://happy-thoughts-api-johanna.herokuapp.com/thoughts

### DIFF
--- a/server.js
+++ b/server.js
@@ -37,6 +37,9 @@ const app = express()
 app.use(cors())
 app.use(bodyParser.json())
 
+//let mongoose use the function
+mongoose.set('useFindAndModify', false);
+
 // Start defining your routes here
 app.get('/', (req, res) => {
   res.send('Hello world')
@@ -62,14 +65,14 @@ app.post('/thoughts', async (req, res) => {
   }
 })
 
-//   app.post('/:id/like', async (req, res) => {
-//   try {
-//     const thought = await Thought.findOneAndUpdate({ _id: req.params.id }, { $inc: { hearts: 1 } })
-//     res.json(thought).status(201);
-//   } catch (err) {
-//     res.status(401).json({ message: 'Heart not added to post', error: err })
-//   }
-// })
+app.post('/:id/like', async (req, res) => {
+  try {
+    const thought = await Thought.findOneAndUpdate({ _id: req.params.id }, { $inc: { hearts: 1 } })
+    res.json(thought).status(201);
+  } catch (err) {
+    res.status(401).json({ message: 'Could not add heart', error: err })
+  }
+})
 
 // Start the server
 app.listen(port, () => {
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -3,14 +3,32 @@ import bodyParser from 'body-parser'
...
+  message: {
+    type: String,
+    required: true,
+    minlenght: 5,
+    maxlength: 140
+  },
+  hearts: {
+    type: Number,
+    default: 0,
+    required: true
```
KarolCislowski, line 21: If you defining the default field, there is no need to make it required, because it will never be empty :)

server.js
```diff
@@ -23,6 +41,37 @@ app.get('/', (req, res) => {
   res.send('Hello world')
 })
 
+app.get('/thoughts', async (req, res) => {
+  const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20).exec()
+  res.json(thoughts)
+})
```
KarolCislowski, line 47: according to documentation `get "/"` need to return 20 thoughts,

server.js
```diff
@@ -23,6 +41,37 @@ app.get('/', (req, res) => {
...
+  const thought = new Thought({ message })
+
+  try {
+    //success
+    const savedThought = await thought.save()
+    res.status(201).json(savedThought)
+  } catch (err) {
+    res.status(400).json({ message: 'could not save thought to the database', error: err.errors })
+  }
+})
```
KarolCislowski, line 62: according to documentation `post "/"` need to create a new one.

server.js
```diff
@@ -23,6 +41,37 @@ app.get('/', (req, res) => {
...
+  try {
+    const savedLike = await Thought.updateOne(
+      { _id: req.params.id }, { $inc: { hearts: 1 } }
+    )
+    res.json(savedLike)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not like post, id not found', error: err.errors })
+  }
+})
+
```
KarolCislowski, line 74: This one should be `/:thoughtId/like`

server.js
jastrand: True, will fix!

server.js
jastrand: I dont really understand this point? It is returning max 20 thoughts?

server.js
jastrand: Not sure I understand this either?

server.js
jastrand: to make it more RESTful?

server.js
KarolCislowski: You used ` get /thoughts` and the documentation forced you to use ` get /`

server.js
KarolCislowski: To meet the requirements. https://www.notion.so/Project-Happy-Thoughts-API-3dc1710667304987894f954dad265fc9

PR #51, opened 2020-05-20
Project Happy Thoughts API - Mathias T - Review wanted

Nesiht:
Fun to connect it to previous project and really cool see the results when everything is working correctly. Really good project that helps to see the connection between frontend with backend.

[Backend](https://plunta-happy-thought-api.herokuapp.com/)
[Frontend](https://relaxed-shockley-ec05e0.netlify.app/)

### DIFF
--- a/server.js
+++ b/server.js
@@ -54,7 +54,16 @@ app.post('/thoughts', async (req, res) => {
 
 // POST /thoughtId/like
 //This endpoint doesn't require a JSON body. Given a valid thought id in the url, tha api should find that thought and uptade its hearts property to add one heart
-
+app.post('/:thoughtId/like', async (req, res) => {
+  const {thoughtId} = req.params
+  
+  try{
+    const likeThought = await Thought.updateOne({ _id: thoughtId } , { $inc: { hearts: 1} })
+    res.status(201).json(likeThought)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not find Id to update', error:err.errors})
+  }
+})
 
 // Start the server
 app.listen(port, () => {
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -2,28 +2,62 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
+import Thought from './model/Thought'
```
KarolCislowski, line 5: External Model 🥇 👍

server.js
```diff
@@ -2,28 +2,62 @@ import express from 'express'
...
-// Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
+})
+
+app.get('/thoughts', async (req, res) => {
+  const thoughts = await Thought.find().sort({createdAt: 'desc'}).limit(20).exec()
+  res.json(thoughts)
+})
```
KarolCislowski, line 36: according to documentation `get "/"` need to return 20 thoughts,

server.js
```diff
@@ -2,28 +2,62 @@ import express from 'express'
...
+  const { message } = req.body
+  const thought = new Thought({ message })
+
+  try{
+    const savedThought = await thought.save()
+    res.status(201).json(savedThought)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not save thought to the databse' , error:err.errors})
+  }
+})
```
KarolCislowski, line 48: according to documentation `post "/"` need to create a new one.

server.js
Nesiht: Updated.

server.js
Nesiht: Updated.

PR #52, opened 2020-05-20
Malins happy thoughts

malin798:
I renamed the /:thoughtsId/like to /thoughts/:thoughtId/like because I think it is more restful since I already had /thoughts as my base endpoint.

I also chose to do a post request to update the likes since I am not sending any information to the server, if I were to update my previous thought I would use put to send in information to the server and update the previously posted information.

### DIFF
--- a/Models/Models.js
+++ b/Models/Models.js
@@ -6,6 +6,10 @@ export const Thought = mongoose.model('Thought', {
     minlength: 5,
     maxlength: 140
   },
+  name: {
+    type: String,
+    default: "Anonymous"
+  },
   hearts: {
     type: Number,
     default: 0
@@ -14,4 +18,3 @@ export const Thought = mongoose.model('Thought', {
     type: String,
   }
 }, "thoughts")
-//.toISOString()
--- a/Models/Models.js
+++ b/Models/Models.js
@@ -1,0 +1,17 @@
+import mongoose from 'mongoose'
+
+export const Thought = mongoose.model('Thought', {
+  message: {
+    type: String,
+    minlength: 5,
+    maxlength: 140
+  },
+  hearts: {
+    type: Number,
+    default: 0
+  },
+  createdAt: {
+    type: String,
+    default: new Date()
+  }
+})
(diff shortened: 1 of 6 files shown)


malin798:
forgot the link:
https://malins-happy-thoughts-my-api.netlify.app/

Models/Models.js
```diff
@@ -0,0 +1,20 @@
...
+export const Thought = mongoose.model('Thought', {
+  message: {
+    type: String,
+    minlength: 5,
+    maxlength: 140
+  },
+  name: {
+    type: String,
+    default: "Anonymous"
+  },
```
KarolCislowski, line 12: Nice! new field :)

server.js
```diff
@@ -2,28 +2,76 @@ import express from 'express'
...
+    }
+  }
+    
+  try {
+    const thoughts = await Thought.find().limit(20).sort( myFilter ) 
+    res.status(200).json(thoughts)
+  } catch(err) {
+    res.status(404).send(err)
+  }
 })
```
KarolCislowski, line 38: according to documentation `get "/"` need to return 20 thoughts,

server.js
```diff
@@ -2,28 +2,76 @@ import express from 'express'
...
+app.post('/thoughts', async (req, res) => {
+  try {
+    await new Thought({ message: req.body.message, name: req.body.name, createdAt: new Date().toISOString()}).save()
+    return res.status(200).json(req.body)
+  } catch(err) {
+    res.status(404).send(err)
+  }
+  
+})
+
```
KarolCislowski, line 49: according to documentation `post "/"` need to create a new one.

server.js
```diff
@@ -2,28 +2,76 @@ import express from 'express'
...
+app.post('/thoughts/:thoughtId/like', async (req, res) => {
+  try {
+    const { thoughtId } = req.params
+    await Thought.updateOne({ _id: thoughtId }, { $inc: { hearts: 1 }})
+    res.status(200).json("update successful")
+  } catch(err) {
+    res.status(404).send(err)
+  }
+})
```
KarolCislowski, line 58: This one should be `/:thoughtId/like`

malin798:
Ok, I changed it!

PR #55, opened 2020-05-20
Happy Thoughts API - Anna G

annaelsamaria:
This was fun! I had some issues with the name not printing out my default value but finally, it worked.

https://anna-happythoughts.herokuapp.com/
https://anna-happythoughts.netlify.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -17,7 +17,6 @@ const Thought = mongoose.model('Thought', {
   },
   name: {
     type: String,
-    minlength: 2,
     maxlength: 50,
     default: "Anonymus"
   },
--- a/server.js
+++ b/server.js
@@ -2,10 +2,28 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
+import { EventEmitter } from 'events'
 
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -18,11 +41,37 @@ const app = express()
...
+    message: message
+  })
+
+  try {
+    const savedThought = await thought.save()
+    res.status(201).json(savedThought)
+  } catch (err) {
+    res.status(400).json({ message: 'Could not save happy thought to database', error: err.errors })
+  }
+})
```
KarolCislowski, line 61: You created the `Name` field in your model, and in the form, but you do not use it here. Add it, please :)

server.js
```diff
@@ -7,6 +7,29 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+    type: String,
+    required: true,
+    minlength: 5,
+    maxlength: 140
+  },
+  name: {
+    type: String,
+    maxlength: 50,
+    default: "Anonymus"
+  },
```
KarolCislowski, line 22: You create this field but you never used it :(

server.js
annaelsamaria: I got some help from Van to get the anonymous posts to work in my frontend. The solution was to remove the name from this part of the code. So I am using it, I think?

PR #61, opened 2020-05-21
William Jensen Happy Thoughts API

Williamriis:
My frontend: https://festive-heyrovsky-dd1229.netlify.app/

My backend: https://williamjensen-happythoughts.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -78,6 +78,10 @@ app.post('/', async (req, res) => {
       postedBy: req.body.user || 'Anonymous',
       createdAt: Date.now()
     }).save()
+    const PAGE_SIZE = 20;
+    const page = req.query.page || 1
+    const order = req.query.order
+    const myOrder = order === 'mostliked' ? { hearts: -1 } : order === 'oldest' ? { createdAt: 1 } : { createdAt: -1 }
 
     const allThoughts = await Thought.find()
     const thoughts = await Thought.find().sort(myOrder).limit(PAGE_SIZE).skip((page * PAGE_SIZE) - PAGE_SIZE)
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -7,20 +7,110 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+      postedBy: username || 'Anonymous',
+      createdAt: Date.now(),
+      theme: req.body.theme || ''
+    }).save()
+    //I return the whole list of thoughts below because if I only return the new thought and add it to the existing state locally
+    //it allows more than the number of thoughts defined in PAGE_SIZE to be displayed simultaneously. 
+    const PAGE_SIZE = 20;
+    const page = req.query.page || 1
+    const order = req.query.order
+    const myOrder = order === 'mostliked' ? { hearts: -1 } : order === 'oldest' ? { createdAt: 1 } : { createdAt: -1 }
```
KarolCislowski, line 94: If you use functionality more then once you should extract it to separate function and call it as often as you need:)  so pagination and sorting are always good candidates for nice function :) And you can create snippets and use it in many another applications :)

server.js
```diff
@@ -7,20 +7,110 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
-app.get('/', (req, res) => {
-  res.send('Hello world')
+
+
+app.get('/', async (req, res) => {
+  const PAGE_SIZE = 20;
+  const page = req.query.page || 1
+  const order = req.query.order
+  const userTheme = req.query.theme ? { theme: req.query.theme } : {}
+  const myOrder = order === 'mostliked' ? { hearts: -1 } : order === 'oldest' ? { createdAt: 1 } : { createdAt: -1 }
```
KarolCislowski, line 72: Someday you will meet someone who is not so good in math and he/she will eat you for this nested ternaries ;)

PR #70, opened 2020-05-22
week 19, happy thoughts api

Sickan-sys:


server.js
```diff
@@ -15,13 +15,69 @@ const port = process.env.PORT || 8080
...
 
-// Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+const thoughts = mongoose.model('thoughts', { 
+  message: {
+        type: String,
+        required: true,
+        minlenght: [2, "At least 2 characters"],
```
KarolCislowski, line 25: You need to add also the `maxlenght` field

PR #73, opened 2020-06-10
Happy Thoughts API

VickBe:
I don't know why this wasn't handed in...

### DIFF
--- a/server.js
+++ b/server.js
@@ -33,7 +33,7 @@ app.use(bodyParser.json())
 
 
 app.get('/', async (req, res) => {
-  const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20)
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -7,22 +7,70 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+
+app.post('/:thoughtId/like', async (req, res) => {
+  const { thoughtId } = req.params
+  //a thought with a specific id
+  const thought = await Thought.findById(thoughtId)
+
+  if (thought) {
+    //likes
+    thought.hearts++
+    thought.save()
```
HIPPIEKICK, line 67: This could actually be done in one line:
`const thought = await Thought.findOneAndUpdate({ _id: req.params.id }, { $inc: { hearts: 1 } }, { new: true })`

PR #74, opened 2020-06-11
Week 19 - Nathalie Rautell

rautellin:
https://rautellin-happy-thoughts-api.herokuapp.com/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,30 @@
-# Project Happy Thoughts API
+# Happy Thoughts App 
 
-Replace this readme with your own information about your project.
-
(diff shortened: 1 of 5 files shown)


HIPPIEKICK:
Add the link yo your connected frontend as well please :)

From project brief:
_Once you've created your API, you should deploy it, and update your frontend project to use your own API instead of the old Technigo one. The idea is that if you build this API correctly, the only thing you should need to change in the frontend code is the url to the API, to change it from the Technigo one, to the one you deploy._

rautellin:
Here's the frontend link 
https://rautellin-happy-thoughts.netlify.app/

PR #82, opened 2021-01-05
Happy Thoughts API 💌 - Sofia Vaz Sousa

sofiavazs:
Happy New Year!
Really fun to be able to connect my backend to a previous front end project, learned a lot about debugging 😅

[Happy Thoughts Back end](https://happy-thoughts-sofia.herokuapp.com/)
[Happy Thoughts Front end ](https://happythoughts-sofia.netlify.app/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,28 @@
-# Project Happy Thoughts API
+# Happy Thoughts API 💌
+This project's goal was to build an API which includes both GET request endpoints to return data and POST endpoints to create data. For the front end I revisited my previous project _Happy Thoughts_ a twitter like app where you can see other people's happy thoughts, post your own and give a ❤️ to an existing one. 
+You can see the live version here, feel free to leave me some happy thoughts! [_Happy Thoughts_](https://happythoughts-sofia.netlify.app/)
 
-Replace this readme with your own information about your project.
+## Planning & What I learned 🧩
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+- This Express API consists of a **MongoDB stored in Atlas and deployed to Heroku** 
+- My first step in this project was to create the mongoose model for the _Thought_, the model schema consists of three main components: message, hearts and createdAt, with several validation rules embbebed, i.e. _required_. 
+- Used _try_ and _catch_ in all endpoints for error handling. 
 
-## The problem
+## Endpoints 💫
+### / 
+Root: Welcome page 
+### GET /thoughts
+Displays the latest 20 thoughts, sorted by createdAt to show the most recent thoughts first.
+### POST /thoughts
+Endpoint to post and save a new thought to the database. It expects a JSON body with the thought message, like this: { "message": "Happy New Year!" }
+### POST /thoughts/:thoughId/like
+This endpoint doesn't require a JSON body. Given a valid thought id in the URL, the API should find that thought, and update its hearts property to add one heart.
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+## Tech ⚡️
+- MongoDB
+- Mongoose
+- Node.js
+- Express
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,13 +1,28 @@
...
+# Happy Thoughts API 💌
+This project's goal was to build an API which includes both GET request endpoints to return data and POST endpoints to create data. For the front end I revisited my previous project _Happy Thoughts_ a twitter like app where you can see other people's happy thoughts, post your own and give a ❤️ to an existing one. 
+You can see the live version here, feel free to leave me some happy thoughts! [_Happy Thoughts_](https://happythoughts-sofia.netlify.app/)
 
-Replace this readme with your own information about your project.
+## Planning & What I learned 🧩
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+- This Express API consists of a **MongoDB stored in Atlas and deployed to Heroku** 
+- My first step in this project was to create the mongoose model for the _Thought_, the model schema consists of three main components: message, hearts and createdAt, with several validation rules embbebed, i.e. _required_.
```
jespa374, line 8: tiny spelling mistake in embedded 😜

README.md
```diff
@@ -1,13 +1,28 @@
...
 
-## The problem
+## Endpoints 💫
+### / 
+Root: Welcome page 
+### GET /thoughts
+Displays the latest 20 thoughts, sorted by createdAt to show the most recent thoughts first.
+### POST /thoughts
+Endpoint to post and save a new thought to the database. It expects a JSON body with the thought message, like this: { "message": "Happy New Year!" }
+### POST /thoughts/:thoughId/like
```
jespa374, line 18: Edit thoughId to thoughtId

server.js
```diff
@@ -1,29 +1,77 @@
...
+// POST REQUESTS
+app.post('/thoughts', async (req,res) => { 
+  try { 
+    const newThought = await new Thought(req.body).save();
+    res.status(200).json(newThought);
+  } catch (error) { 
+    console.log(error);
+    res.status(400).json({ success:false, error });
+  };
+});
```
jespa374, line 60: At the moment it's possible to post likes for example by adding hearts property in the JSON body. I added 85 hearts by adding that with my message which worked. To prevent this from happening I think you can change line 54 to const newThought = await new Thought( { message }.save(); and adding const { message } = req.body; between line 53 and 54. hope that works  👍   

It also looks like you can reassign createdAt by adding that info in the JSON  body, however the thought is not posted so I guess it's fine. But I think that you could prevent that by the previous code where you only accept message property from whatever the user sends in the request.

jespa374 (changes requested):
Great job Sofia! 🥇  Clean code and the frontend and the backend works as expected.  The only thing that I would suggest you look at is the possibility to send number of hearts when you send a message. This should not be reassignable but hope my code suggestion helps you solve that issue. Well done with everything else ! 😃

README.md
sofiavazs: Thank you! fixed!

server.js
sofiavazs: Thank you for spotting that! Fixed and now only 1 heart can be sent! which is sad, I want more 85 hearts !! 🙈♥️

server.js
```diff
@@ -1,29 +1,84 @@
...
+     default: Date.now
+   }
+ });
 
-// Start defining your routes here
+// ROUTES
 app.get('/', (req, res) => {
-  res.send('Hello world')
-})
+  res.send('Hello and Welcome to the Happy thoughts API, where feeling blue is not allowed! :)');
```
KarolCislowski, line 42: I'm blue
Da ba dee da ba di...

Sorry I couldn't help myself 😸

KarolCislowski (approved):
Great work 😺

server.js
sofiavazs: da ba di da bi daaaa
😂 loved it!

PR #84, opened 2021-01-06
Project Happy Thoughts API - Vanessa Sue Smith

VanessaSue27:
This was another super fun project since we could work both in the Backend and Frontend for it 😍
I managed to complete the blue requirements fast, so I decided to challenge myself with some stretch goals that I found interesting.

(It was so handy that the paths for the endpoints were already written for us 😋, writing RESTful paths is a whole science!)

I added an extra feature so the users can sign their Thoughts with a "username". Implementing that in the Frontend took longer than doing it in the Backend, but it was a great practice to learn how to add more fields to the object being sent to the API and effectively adding that to the model saved to the database. I enjoyed doing that 👌

I have also implemented Pagination on the main GET endpoint. I had done pagination before with vanilla JS, so it was great learning to do that with mongoose, which turned out to be simpler than I expected.
I also implemented that on my Frontend, so now there are navigation buttons to browse between recent and older thoughts 😁💕

Frontend updated with the new features is here: https://happy-vane-thoughts.netlify.app/

Really enjoying these projects where I get to work with both ends!
// Vanessa 🍉

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ For this to work, I created an API in the Backend using **Express and MongoDB** 
 - The setup for this Express API consists of a **MongoDB stored in Atlas and deployed to Heroku** 💥
 - In order to create the endpoints and manipulate the data, I'm working with **mongoose**. I have one main model for 'Thought'. This schema includes various validation rules in order to control that the information saved to the database is clean. Super good learning how to validate the data on this project: I used properties like required, min and max length.
 - I also learned how to implement **Pagination** using mongoose, which I use in my main GET endpoint which shows the 20 most recent Thoughts. This endpoint shows 20 Thoughts at a time (using limit()), starting with the most recent ones (using sort()).
-- On my main POST endpoint to add a new Thought to the database, I managed to implement a new feature in which the user can **sign the tought**. So I'm sending both "message" and "username" properties to the databse when creating a new entry. If the username field is left empty, it will be automatically created as "Anonymous".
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,13 +1,39 @@
...
+
+If the input is valid (message must be between 5 - 140 characters), the thought will be saved and you will get a response showing the Thought object that was just created. If the input is invalid, you'll get an error message explaining what went wrong.
+
+### POST /thoughts/:thoughtId/like
+This endpoint does not require a JSON body. Given a valid thought id in the URL, the API should find that thought and update its hearts property to add one heart. If the id is invalid, an error message will show.
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+💗 Happy Thoughts API 💗 is live - powered by Heroku here: https://vane-happy-thoughts.herokuapp.com/
```
KarolCislowski, line 39: Great documentation 👍🏻

server.js
```diff
@@ -1,29 +1,106 @@
...
+  const { page = 1, limit = 20 } = req.query;
+
+  try {
+    // execute query with page and limit values
+    const thoughts = await Thought.find()
+      .limit(limit * 1)
+      .skip((page - 1) * limit)
+      .sort({ createdAt: 'desc' })
+      .exec();
+
```
KarolCislowski, line 58: Mongoose pagination ❤️

KarolCislowski (approved):
Awesome project 🥇

PR #85, opened 2021-01-06
Project Happy Thoughts API - Emelie Svensson

emeliesv:
This week I've been very tired and so I decided to only reach for the blue requirements. 
I'm satisfied with how the database turned out, and if I had more energy and time I would have added the name-function for the red level and also pagination from the black level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,3 +29,4 @@ The POST /thoughts/:id/like endpoint return a 404 if the thought is not found.
 
 ## View it live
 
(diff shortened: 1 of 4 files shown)


idsintehittapa (approved):
Hey Emelie! 
Nice job! Especially if you felt a bit low that week. 

I’ve tried out your happy though and it works beautifully! I sent in a new thought and liked an old one, the database updated as it should with the post request.

KarolCislowski (approved):
Nice work 👍🏻

PR #87, opened 2021-01-07
Week 19: Happy thoughts API

monkeybusiness101:
https://mongo-thoughts-api.herokuapp.com/
https://sleepy-carson-4649f3.netlify.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -66,7 +66,6 @@ app.post('/thoughts/:id/like', async (req, res) => {
   } else {
     res.status(400).json({message: "can not find thought"})
   }
-
 })
 
 // Start the server
--- a/server.js
+++ b/server.js
@@ -2,10 +2,28 @@ import express from 'express'
 import bodyParser from 'body-parser'
 import cors from 'cors'
 import mongoose from 'mongoose'
+import { restart } from 'nodemon'
 
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
+
+const Thought = new mongoose.model('Thought', {
+  message: {
+    type: String,
+    required: true,
+    minlength: 5,
+    maxlength: 140
+  },
+  hearts: {
+    type: Number,
+    default: 0
+  },
+  createdAt: {
+    type: Date,
+    default: Date.now
+  }
+})
 
 // Defines the port the app will run on. Defaults to 8080, but can be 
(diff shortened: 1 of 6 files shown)


server.js
```diff
@@ -2,11 +2,31 @@ import express from 'express'
...
+    type: Number,
+    default: 0
+  },
+  createdAt: {
+    type: Date,
+    default: Date.now
+  }
+})
+
 // Defines the port the app will run on. Defaults to 8080, but can be
```
pwangy, line 30: Should these comments be removed?

server.js
```diff
@@ -19,8 +39,33 @@ app.use(cors())
...
+
+app.post('/thoughts/:id/like', async (req, res) => {
+  const { id } = req.params
+  const findThought = await Thought.findOne({_id: id})
+
+  if (findThought) {
+    await Thought.updateOne({_id: id}, {$inc : {hearts: 1}})
+    res.status(201).json({message: `added like to ${id}`})
+  } else {
+    res.status(400).json({message: "can not find thought"})
```
pwangy, line 67: Consistency check here: Why use single quotes on line 65 and double qoutes on 67?

server.js
pwangy: and on lines: 37, 41, and 71

server.js
```diff
@@ -19,8 +39,33 @@ app.use(cors())
...
+
+  try {
+    const saveThought = await thoughts.save()
+    res.status(201).json(saveThought)
+  } catch (err) {
+    res.status(400).json({message: 'could not save task to db', error:err.errors})
+  }
+})
+
+app.post('/thoughts/:id/like', async (req, res) => {
```
pwangy, line 59: consider changing this one to PUT instead of POST. 

"POST is always for creating a resource ( does not matter if it was duplicated ) PUT is for checking if resource is exists then update , else create new resource. PATCH is always for update a resource."

pwangy (changes requested):
Hi Max! I'm impressed that you took the time over the holidays to do some advanced work on your own. Great initiative!

I'm making a small request to change the "like" request from a Post to a Put. 

The other comments are optional changes. And one day, if you want to I would like to encourage you to update the frontend so that it gives your users feedback when they try to submit a message which is too long. Or some kind of indicator of what the character limit might be on the app. 

Overall, lovely work and clean code. It was a pleasure to review.

server.js
KarolCislowski: Your way of thinking is right, but documentation said that it must be 'POST'.

KarolCislowski (approved):
Great work 👍🏻

PR #91, opened 2021-01-07
Pull request for happy thoughts api

Katsiez:


server.js
```diff
@@ -1,28 +1,127 @@
...
+        createdAt: -1,
+      };
+    }
+  };
+
+  const thoughts = await Thought.find()
+    .sort(sortAllThoughts(sort))
+    .limit(thoughtsPerPage)
+    .skip(skip)
+    .exec();
```
liza-minican, line 87: Not sure what this code does as I don´t see you using this const later on in the code.

liza-minican (commented):
Hey Kat,

I always admire your desire to learn more and that you are never scared of implementing new things!!!! I've got inspired by you many times!! Keep up your curiosity and good work!!! You did a great job with the project. 
https://happy-thoughts-kat.netlify.app/ this page doesn't show your form. I checked your code in the happy thoughts project and it looks like you connected the links correctly. Not sure why it is happening, maybe there is an issue with connection to atlas. Hopefully our code coaches can help you fix that. 
Good luck with everything!! Be proud!

KarolCislowski (changes requested):
Great work with the API, but your frontend do not show anything. I looked inside and I have some idea how to fix it:
in app.js in line 45:
  `{thoughts.map(thought => (` 
change to:
 ` {thoughts?.map(thought => (`
 
this `?` makes map start only if your array is not empty, so it will not crash before `useEffect()` will start.

Katsiez:
Thanks for your comment, Karol! I knew the frontend wasn't connecting for some odd reason, made the change you suggested and I am still experiencing the same issue, sadly.

KarolCislowski:
Your api is returning 
```
{
  "pages": 1,
  "thoughts": [...]
 }
  ```
  So you need to access `thoughts` inside your `json` in `useEffect`:
  `.then(json => setThoughts(json.**thoughts**))`

PR #101, opened 2021-01-08
Handing in project 19

HenrikeW:
It was a really nice experience creating the backend for an already existing frontend!

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,7 +1,7 @@
-import express from 'express'
-import bodyParser from 'body-parser'
-import cors from 'cors'
-import mongoose from 'mongoose'
+import express from 'express';
+import bodyParser from 'body-parser';
+import cors from 'cors';
+import mongoose from 'mongoose';
 
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,29 +1,76 @@
...
+    const savedThought = await thought.save();
+    res.status(201).json(savedThought);
+  } catch (err) {
+    res.status(400).json({ message: 'Could not save thought', error: err.errors });
+  }
+});
+
+// update the number of likes on a thought using a PUT request
+app.put('/thoughts/:id/like', async (req, res) => {
+  try {
```
IngelaL, line 64: Interesting that you used PUT here. I didn't think about it myself but it makes sense what you mention about it in the read me file,

IngelaL (approved):
Hi Henrike! Your project looks great and you are very approved. 🎉 I have just a few suggestions on  what you could do, if you feel like it :)  //Ingela

server.js
```diff
@@ -1,29 +1,76 @@
...
+    const savedThought = await thought.save();
+    res.status(201).json(savedThought);
+  } catch (err) {
+    res.status(400).json({ message: 'Could not save thought', error: err.errors });
+  }
+});
+
+// update the number of likes on a thought using a PUT request
+app.put('/thoughts/:id/like', async (req, res) => {
```
KarolCislowski, line 63: `PUT` is a very good type of request for updating the data, but the documentation said that you have to use `POST`. Please change it. 😸

KarolCislowski (changes requested):
It is a very nice project, but you need to change your endpoint to make it aligned with the documentation.

PR #106, opened 2021-01-09
Week 19 - project-happy-thoughts-api - Nadia Mazzoni

Bluepoisson:


server.js
```diff
@@ -3,27 +3,100 @@ import bodyParser from 'body-parser'
...
+  } else {
+    res.status(503).json({ error: 'Service Unavailable' })
+  }
+})
+
+
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
 })
```
LillanEriksson, line 63: Nice that you listed all the endpoints!

server.js
```diff
@@ -3,27 +3,100 @@ import bodyParser from 'body-parser'
...
+    res.status(400).json({ message: 'Bad request. Could not save the happy thought to the database', error: err.errors });
+  }
+});
+
+//? put instead of post because updating existing hearts data, increments by one
+app.put('/thoughts/:thoughtId/like', async (req,res) => {
+  const { thoughtId } = req.params
+  
+  try {
+  const onMessageLiked = await Thought.findOneAndUpdate({ _id: thoughtId }, {$inc: { hearts: 1 }})
```
LillanEriksson, line 93: Since you wrote that you had problems with this, I can share how I did.
I used updateOne instead of findOneAndupdate, and instead of id: thoughtId I used id: req.params.thoughtId.

server.js
```diff
@@ -3,27 +3,100 @@ import bodyParser from 'body-parser'
...
+    const savedThought = await thought.save();
+    res.status(200).json(savedThought);
+  } catch (err) {
+    res.status(400).json({ message: 'Bad request. Could not save the happy thought to the database', error: err.errors });
+  }
+});
+
+//? put instead of post because updating existing hearts data, increments by one
+app.put('/thoughts/:thoughtId/like', async (req,res) => {
+  const { thoughtId } = req.params
```
LillanEriksson, line 90: Maybe try req.params.thoughtId?

server.js
```diff
@@ -3,27 +3,100 @@ import bodyParser from 'body-parser'
...
+    res.status(200).json(onMessageLiked)
+  } catch (err) {
+    res.status(404).json({ message: 'Could not find ThoughtId, update not possible', errors: err.errors});
+  };
+});
+
 app.listen(port, () => {
   console.log(`Server running on http://localhost:${port}`)
-})
+})
```
LillanEriksson, line 102: Good work! Also tried it out in the front-end part and left a little message for you :) I will set this to requested changes so you can fix the likes-functionality. Ping me on Slack if you need help :)

LillanEriksson (changes requested):
Nice work! Just get the like-functionality to work, then blue level is top notch.

server.js
Bluepoisson: Thank you :) The problem is solved. I forgot to change the method to PUT in the fetch in the frontend and needed to change method in the backend to findOneByIdAndUpdate.

server.js
```diff
@@ -3,27 +3,99 @@ import bodyParser from 'body-parser'
...
+  try {
+    const savedThought = await thought.save();
+    res.status(200).json(savedThought);
+  } catch (err) {
+    res.status(400).json({ message: 'Bad request. Could not save the happy thought to the database', error: err.errors });
+  }
+});
+
+//? put instead of post because updating existing hearts data, increments by one
+app.put('/thoughts/:thoughtId/like', async (req,res) => {
```
KarolCislowski, line 89: `PUT` is a good request to update your data, but the documentation required that it must be `POST`. Change it, please 😺

KarolCislowski (changes requested):
Please change your request type to fit the documentation.

server.js
Bluepoisson: Changed in backend and frontend :)

KarolCislowski (approved):
This is a very nice and clean code 🦄

PR #110, opened 2021-01-09
Happy thoughts API - Cornelia 

Corneliaryas:
Live: https://sharing-happy-thoughts.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -7,7 +7,24 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
 mongoose.Promise = Promise
 
-// Defines the port the app will run on. Defaults to 8080, but can be 
+const Thought = mongoose.model('Thought', {
+  message: {
+    type: String,
+    required: true,
+    minlength: 5,
+    maxlength: 140
+  },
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -7,7 +7,33 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+    default: Date.now
+  },
+  username: {
+    type: String,
+    maxlength: 50
+  },
+  tags: {
+    type: Array,
+    of: String
+  }
```
KarolCislowski, line 32: Great that you did this extra step 👍🏻

server.js
```diff
@@ -23,6 +49,47 @@ app.get('/', (req, res) => {
...
+  const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20)
+  if (thoughts.length > 0){ 
+    res.json(thoughts)
+  } else {
+    res.json({ message: "No thoughts found" })
+  }
+
+})
+
+app.get('/thoughts/:tags', async (req, res) => {
```
KarolCislowski, line 62: A better way to achieve it is to use queries, or change the endpoint to `/thoughts/tag/:tag`. Because intuition says that after `/thoughts` should be `:id`

KarolCislowski (approved):
Awesome project 🥇

PR #115, opened 2021-01-10
Sandrine's "Happy Thoughts by M API" 

MindstormingAB:
Link to my deployed project: https://happy-thoughts-by-m.herokuapp.com/
Link to the frontend project: https://happy-tweets-by-m.netlify.app/

My reflections on how this week's project turned out: The blue requirements felt fairly accessible this week and it was so rewarding to have a "ready for use" frontend code to connect to this API. I couldn't leave this project without experimenting more with Mongoose's aggregate functions on backend and useState and useEffect on frontend :-)

Things I'd like to have clarified or explained in more detail: --


monkeybusiness101 (changes requested):
Hi Sandrine! 

I have looked through your code and want to say good job! You tried some things that feels advanced like the aggregate of categories or authors. As we have talked about on slack, you might have an infinite loop here that needs to be resolved. 

I looked through the blue reqs and you have passed those. Please let me know when you have fixed the loop and I will approve you (unless you want more feedback from code coach)

Blue level:
- Your API should implement the routes exactly as documented above.
- I looked through the reqs and it looks good! I did one test where I tried to post a thought with several likes. You do not save the likes and default to 0 👍 
- Your `GET /thoughts` endpoint should only return 20 results, ordered by `createdAt` in descending order.
- You are sorting on descending and limit 20 on row 61 👍 
- Your API should validate user input and return appropriate errors if the input is invalid.
- I do not know how extensive the validation should be but you have added validation in the mongoose model which I believe is sufficient for this project 👍 
- In the `POST /thoughts` endpoint to create a new thought, if the input was invalid and the API is returning errors, it should set the response status to `400` (bad request).
-  Checked in Postman and you provide 400 with mongoose error if post body is not in the correct format 👍 
- The endpoint to add hearts to a thought should return an appropriate error if the thought was not found.
- Checked in Postman and you provide 400 if the id does not exist 👍

MindstormingAB:
Hi Max @monkeybusiness101!
Thank you for making aware of these loops!
I reviewed the frontend and adjusted the useEffects associated to the filters so they will be triggered by `toughts` instead of `categories`/`authors`. The loops seem to be gone now. :-)

server.js
```diff
@@ -1,29 +1,121 @@
...
+      { $group: { _id: "$author" } },
+      { $sort: { _id: 1 } },
+      { $out: "authors" }
+    ]);
+    const authors = await Author.find().exec();
+    res.status(200).json(authors);
+  } catch (err) {
+    res.status(400).json({ err });
+  };
+});
```
KarolCislowski, line 116: Sweet aggregations! 🎆

KarolCislowski (approved):
This is a Top Tier project! 🥰

PR #117, opened 2021-01-10
Happiest happy thougths - Evelina Berg

evelinabe:
It was sad to let the old API with everyone's thoughts go, but also cool to have done everything by yourself. 

View it live: 
https://happiest-happy-thoughts.netlify.app/ 
https://happiest-happy-thoughts.herokuapp.com/


server/server.js
```diff
@@ -0,0 +1,115 @@
...
+app.use(bodyParser.json());
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next();
+  } else {
+    res
+      .status(500)
+      .json({ error: "Service unavailable" });
+  }
+});
```
karinnordkvist, line 30: Wow! Learned something new here. Nice way of handling an error with the service! 🤩

server/server.js
```diff
@@ -0,0 +1,115 @@
+import express from "express";
+import bodyParser from "body-parser";
+import cors from "cors";
+import mongoose from "mongoose";
+import dotenv from "dotenv";
+
+dotenv.config();
```
karinnordkvist, line 7: What is this being used for? 🤓

server/server.js
```diff
@@ -0,0 +1,115 @@
...
+        message: "Thought must be longer than 5 characters, please try again.",
+      },
+      {
+        validator: (thoughtValue) => {
+          return thoughtValue.length < 140;
+        },
+        message:
+          "Thought must be shorter than 140 characters, please try again",
+      },
+    ],
```
karinnordkvist, line 50: These validations could be done simply by adding a max-length/min-length and specifying the message after it 😊, like so: 
`minlength: [5, "Thought must be longer than 5 characters, please try again."],`

karinnordkvist: Oh I noticed now that you did exactly this for the required part above. You can also do it for these two. 👍

server/server.js
```diff
@@ -0,0 +1,115 @@
...
+if (process.env.RESET_DATABASE) {
+  const resetDatabase = async () => {
+    await Thought.deleteMany();
+  };
+  resetDatabase();
+}
+const listEndpoints = require('express-list-endpoints');
+app.get("/", (req, res) => {
+  res.send(listEndpoints(app));
+});
```
karinnordkvist, line 71: Nice! 🌟

server/server.js
```diff
@@ -0,0 +1,115 @@
...
+    console.log(`GET /thoughts`);
+    const thoughtsList = await Thought.find()
+      .sort({ createdAt: "desc" })
+      .limit(20)
+      .exec();
+    res.status(200).json(thoughtsList);
+  } catch (error) {
+    res
+      .status(404)
+      .json({ message: `Couldn't find any thoughts`, error });
```
karinnordkvist, line 84: I was told that when sending a json-response, you should format it with a key and a value, so for the error you could write like `error: error`. 😊

server/server.js
```diff
@@ -0,0 +1,115 @@
...
+    res
+      .status(400)
+      .json({ message: "Could not save the thought to the database", error });
+  }
+});
+
+app.post("/thoughts/:id/likes", async (req, res) => {
+  try {
+    await Thought.updateOne({ _id: req.params.id }, { $inc: { hearts: 1 } });
+    res.status(201).json();
```
karinnordkvist, line 105: Not sure what the response will be when the json is empty? 👀 Or maybe it's just me being stupid..

karinnordkvist (approved):
Really nice job Evelina! You managed to do some really nice error handling, and it was inspirational to see that you added some handling for the service in whole as well. 👍 Keep up the good work as always! 💕

server/server.js
```diff
@@ -0,0 +1,115 @@
...
+  },
+});
+
+if (process.env.RESET_DATABASE) {
+  const resetDatabase = async () => {
+    await Thought.deleteMany();
+  };
+  resetDatabase();
+}
+const listEndpoints = require('express-list-endpoints');
```
KarolCislowski, line 68: You shouldn't mix  2 methods of importing external modules. Choose between `import` and `require` syntax and be consistent. 😸

KarolCislowski (approved):
Very nice project 💯

PR #121, opened 2021-01-10
Happy-API

lilihorvath:


Snabbina:
Hey Lili, hope you are good! Im supposed to review your project, but the Heroku link does not seem to work... de deployed link should look something like this "https://name-oftheproject-api.herokuapp.com/" and plz also add the deployed link to the frontend so I can check that out too for you! /Have a nice evening! /S

server.js
```diff
@@ -7,23 +7,63 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+  createdAt: {
+    type: Date,
+    default: Date.now
+  }
 })
 
-// Start the server
-app.listen(port, () => {
-  console.log(`Server running on http://localhost:${port}`)
+app.get('/', async (req, res) => {
```
KarolCislowski, line 37: this endpoint should be `GET /thoughts` 😸

server.js
```diff
@@ -7,23 +7,63 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+  const thoughts = await Thought.find().sort({createdAt: 'desc'}).limit(20).exec()
+  res.json(thoughts)
 })
+
+app.post('/', async (req, res) => {
```
KarolCislowski, line 42: this endpoint should be `POST /thoughts` 😸

server.js
```diff
@@ -7,23 +7,63 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+  message: {
+    type: String,
+    required: true, 
+    minLength: 5,
+    maxLength: 140,
+  },
+  username: {
+    type: String, 
+    maxLength: 20,
+  },
```
KarolCislowski, line 26: you created this field in the model, but you never used it. why? 🤔

KarolCislowski (changes requested):
Please change wrong endpoints and add a link to the hosted frontend. 😺

PR #122, opened 2021-01-10
Project Happy Thoughts API Joel

JHarde:


server.js
```diff
@@ -1,29 +1,92 @@
...
-import mongoose from 'mongoose'
+import express from "express";
+import bodyParser from "body-parser";
+import cors from "cors";
+import mongoose from "mongoose";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
-mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true })
-mongoose.Promise = Promise
+const listEndpoints = require("express-list-endpoints");
```
Corneliaryas, line 6: Super nice to list the endpoints!

server.js
```diff
@@ -1,29 +1,92 @@
...
 // Add middlewares to enable cors and json body parsing
-app.use(cors())
-app.use(bodyParser.json())
+app.use(cors());
+app.use(bodyParser.json());
+
+const Thought = mongoose.model("Thought", {
+  message: {
+    type: String,
+    required: [true, "Message is required"],
```
Corneliaryas, line 26: Nice with error message 👍

server.js
```diff
@@ -1,29 +1,92 @@
...
+app.get("/", (req, res) => {
+  res.send(listEndpoints(app));
+});
+
+app.get("/thoughts", async (req, res) => {
+  const thoughts = await Thought.find().sort({ createdAt: "desc" }).limit(20);
+  if (thoughts.length > 0) {
+    res.json(thoughts);
+  } else {
+    res.status(400).json({ error: "No thoughts found" });
```
Corneliaryas, line 55: 👍

server.js
```diff
@@ -1,29 +1,92 @@
...
+  }
+});
+
+app.post("/thoughts", async (req, res) => {
+  try {
+    const { message, username } = req.body;
+    const thought = await new Thought({
+      message,
+      username: username || "Anonymous",
```
Corneliaryas, line 64: Nice bonus with username! How does the "Anonymous" text work here, do you need a clearer conditional (username ? username : "Anonymous") or does it works as intended like this?

Corneliaryas (approved):
Great work with the Happy thoughts API Joel! Did not find a lot to comment on, the site works well as intended and you have a clear, easy-to-follow code! Nice bonus with the extra field of username! 😊🎉

server.js
KarolCislowski: Nice functionality but... You shouldn't mix  2 methods of importing external modules. Choose between `import` and `require` syntax and be consistent. 😸

server.js
```diff
@@ -1,29 +1,92 @@
...
+    type: Number,
+    default: 0,
+  },
+  createdAt: {
+    type: Date,
+    default: Date.now,
+  },
+  username: {
+    type: String,
+    default: "Anonymous",
```
KarolCislowski, line 41: Nice that you created an extra field!

KarolCislowski (approved):
It is a very nice project 😸

PR #127, opened 2021-01-10
Peggy's Happy Thoughts API

pwangy:
Link to deployed site: https://happy-thoughts-api-pwangy.herokuapp.com/

Reflections: It was difficult to get back into the flow of things this week after the break. I needed the break, but I had a hard time settling down to focus on this one. I'll be coming back to it to keep tinkering.

Questions: Nothing at the moment.


server.js
```diff
@@ -7,23 +7,60 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+    res.status(400).json({message:'Could not save thought', errors: err.errors})
+  }
+})
+
+// Add a heart to a thought
+app.put('/thoughts/:id/like', async (req,res) => {
+  try {
+    const thought = await Thought.findById(req.params.id)
+    await Thought.updateOne({_id: thought._id}, {hearts: thought.hearts + 1})
+    res.status(200).json(thought)
```
HenrikeW, line 58: One thing I noticed when trying this endpoint in Postman: 
When I try to like a thought with, say, 0 likes I can do that (which is good! 😄 ) But the JSON-response I'm getting back still says hearts: 0, even though when I do a GET request to /thoughts I can see that my like has reached the database. That's because in line 58 you send back the original thought from line 56 and not the updated one.

It would be nice to get the updated thought as a response and I think you could do it like this: 
```suggestion
    const thought = await Thought.findById(req.params.id)
    const updatedThought = await Thought.updateOne({_id: thought._id}, {hearts: thought.hearts + 1})
    res.status(200).json(updatedThought)
```

package.json
```diff
@@ -9,18 +9,18 @@
...
-    "mongoose": "^5.8.10",
-    "nodemon": "^2.0.2"
+    "@babel/core": "*",
+    "@babel/node": "*",
+    "@babel/preset-env": "*",
+    "body-parser": "*",
+    "cors": "*",
+    "express": "*",
+    "mongoose": "*",
+    "nodemon": "*"
```
HenrikeW, line 19: So nice with the "*" solution!! 💯

README.md
```diff
@@ -1,13 +1,39 @@
...
+
+## Completed Requirements
+🔵  Blue Level
+- The API should be deployed to Heroku or similar hosting service.
+- The database should be deployed using mongo cloud or similar.
+- The API should implement the routes exactly as documented in the brief (A copy of the brief can be found in the production folder as a .pdf).
+- The `GET /thoughts` endpoint should only return 20 results, ordered by `createdAt` in descending order.
+- The API should validate user input and return appropriate errors if the input is invalid.
+- In the `POST /thoughts` endpoint to create a new thought, if the input was invalid and the API is returning errors, it should set the response status to `400` (bad request).
+- The endpoint to add hearts to a thought should return an appropriate error if the thought was not found.
```
HenrikeW, line 24: Really good to include this into the readme! 👍

HenrikeW (approved):
Not much to comment here, in a good way! Your code is straight forward and easy to follow and as far as I can see from testing it in the browser and in Postman, it does what it's supposed to do. Great job, Peggy! 🥇 
I noticed one minor issue in a JSON response that shouldn't be a problem for the user but might be for a developer. I think it could easily be fixed with just a small change in the code, see my suggestion. 
I couldn't see if your code works together with the frontend from the old project but as that wasn't part of the requirements technically, I think I can approve this. Well done! 🎉

server.js
```diff
@@ -7,23 +7,60 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+  } catch (err) {
+    res.status(400).json({message:'Could not save thought', errors: err.errors})
+  }
+})
+
+// Add a heart to a thought
+app.put('/thoughts/:id/like', async (req,res) => {
+  try {
+    const thought = await Thought.findById(req.params.id)
+    await Thought.updateOne({_id: thought._id}, {hearts: thought.hearts + 1})
```
KarolCislowski, line 57: you should assign `const thought` to a `updateOne` line, and remove logic from line 56 because it does nothing you are using.

server.js
KarolCislowski: Eagle Eye @HenrikeW 🥇

KarolCislowski (changes requested):
Please fix the issue in line 56 and add a link to the frontend to the README for testing.

pwangy:
@KarolCislowski Thanks for the nudge. I've made the requested changes and connected the backend to a clone of the original frontend so I can have both versions.

KarolCislowski (approved):
This is nice and clean code 🌈 Good job 🦄

PR #132, opened 2021-01-13
W. 19 Claudia 

ckjerrgren:
Not finished, but handed in so Vanessa can do a code review on what I've done so far.

### DIFF
--- a/server.js
+++ b/server.js
@@ -35,16 +35,16 @@ app.use(bodyParser.json())
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello Claudia')
+  res.send('Hello Teachers')
 })
 
-// 1. GET endpoint returning 20 thoughts in descending order from created date.
+// GET endpoint returning 20 thoughts in descending order from created date.
 app.get('/thoughts', async (req, res) => {
   const thoughts = await Thought.find().sort({createdAt: 'desc'}).limit(20).exec()
   res.json(thoughts)
 })
 
-// 2. POST to database 
+//  POST to database 
 app.post('/thoughts', async (req, res) => {
   //Retrive the information sent by the client to our API endpoint 
   try {
@@ -57,7 +57,7 @@ app.post('/thoughts', async (req, res) => {
 
 
 
-// 3. POST hearts (for likes)
+// POST hearts (for likes)
 app.post('/thoughts/:thoughtId/heart', async (req, res) => {
   const { thoughtId } = req.params
   try {
--- a/server.js
+++ b/server.js
@@ -11,8 +11,8 @@ const Thought = mongoose.model('Thought', {
     text: {
       type: String,
       required: true,
-      minLength: 5,
-      maxLength: 140,
+      minlength: 5,
+      maxlength: 140,
     },
     heart: {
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -7,10 +7,25 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+const Thought = mongoose.model('Thought', {
+    text: {
+      type: String,
+      required: true,
+      minlength: 5,
+      maxlength: 140,
+    },
+    heart: {
+      type: Number,
+      default: () => 0
```
VanessaSue27, line 19: I think here default can be just `default: 0` instead of using the `() => 0` syntax

server.js
```diff
@@ -7,10 +7,25 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
...
+    },
+    createdAt: {
+      type: Date,
+      default: Date.now,
+    },
+});
```
VanessaSue27, line 25: Semi colon here 👀

server.js
```diff
@@ -20,7 +35,27 @@ app.use(bodyParser.json())
 
 // Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Hello Claudia')
+})
+
+// endpoint returning 20 thoughts in descending order from created date.
+app.get('/thoughts', async (req, res) => {
+  const thoughts = await Thought.find().sort({createdAt: 'desc'}).limit(20).exec();
```
VanessaSue27, line 43: Lines 43, 44, 45 and 52: Semi colons consistency, in some places I see you use them, but it seems most the code is without them? 👀

server.js
```diff
@@ -20,7 +35,27 @@ app.use(bodyParser.json())
...
+  res.json(thoughts);
```
VanessaSue27, line 44: This endpoint is looking good to me 👌

server.js
```diff
@@ -20,7 +35,27 @@ app.use(bodyParser.json())
...
+});
+
+app.post('/thoughts', async (req, res) => {
+  //Retrive the information sent by the client to our API endpoint 
+  const {text, complete} = req.body;
```
VanessaSue27, line 49: Ok so I see this endpoint is still under construction 🚧
I understand this "text" is the text we will get from the client (what the use has written in the text field), but not sure what "complete" is.
Hard to tell since the endpoint is not finished, so maybe it becomes more understandable when it's done 🙃

VanessaSue27 (commented):
Hello Claudia! ☺

Not so much for me to review since I see you are still working on this project, but I looked at it as thoroughly as I could 💪
You seem to be going in the right direction, so hope it all ties up nicely at the end 💌

Remember to connect the Backend to the Frontend once you are done, so you can test if everything is working smoothly 😁

I can always try and have a look later on once you are finished, but for now I just leave these comments, but will not Approve or Request Changes.

// Vanessa 🍉

KarolCislowski (changes requested):
Please finish both `POST` endpoints and add both links(frontend and backend) into README file.

ckjerrgren:
Finished

PR #143, opened 2021-05-20
Siri Hemsley - Happy thoughts API

sirimatsdotter:
This week felt so much easier than last week and I didn't have any problems completing the blue requirements. I didn't prioritise moving further though since I spent some time on my Final project pitch instead. 

Find my API here: https://happy-thoughts-siri.herokuapp.com/
And the front end here: https://awesome-murdock-ef4ca9.netlify.app/


server.js
```diff
@@ -1,29 +1,114 @@
...
-// Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
+})
+
+// End point to POST a new thought
+app.post('/thoughts', async (req, res) => {
+  try {
+    const savedThought = await new Thought(req.body).save()
```
annalenksjo, line 50: Here you have to write (req.body.message) to prevent the possibility to send values for hearts and/or Date, or some other method to prevent this :-)

annalenksjo (changes requested):
Hej Siri,
Nice and well-structured project as always! Great job that you connected it with your front-end as well 👍 
I like that you have some comments in your project, it makes it easier to follow, and I guess for your self as well if you come back to the project later on :-) 

All the endpoints seems so be working fine, and the errormessages too. One suggestion is that you organize your endpoints by method in server.js. I think Max mentioned this in one of the lectures as sort of a standard. 

There is one thing you missed, which is to prevent the possibility to post hearts and Date in the message-post request :-). See my comment in the code. Otherwise the project meets all of the blue requirements. Good job! 🌷

server.js
```diff
@@ -1,29 +1,114 @@
...
 const app = express()
 
-// Add middlewares to enable cors and json body parsing
+const thoughtSchema = new mongoose.Schema({
+  message: {
+    type: String,
+    required: [true, "Message is required"],
+    unique: [true, "Message has to be unique"],
+    minlength: [5, "Message has to be at least 5 characters"],
+    maxlength: [140, "Message cannot exceed 140 characters"]
```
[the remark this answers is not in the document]
KarolCislowski: Really nice and clean method 😻

server.js
```diff
@@ -1,29 +1,114 @@
...
-//   PORT=9000 npm start
 const port = process.env.PORT || 8080
 const app = express()
 
-// Add middlewares to enable cors and json body parsing
+const thoughtSchema = new mongoose.Schema({
+  message: {
+    type: String,
+    required: [true, "Message is required"],
+    unique: [true, "Message has to be unique"],
```
KarolCislowski, line 24: Why you used `unique` here? people can't be happy for the same reasons? 😭

server.js
```diff
@@ -1,29 +1,114 @@
...
+app.use(express.json())
 
-// Start defining your routes here
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
+})
+
+// End point to POST a new thought
+app.post('/thoughts', async (req, res) => {
```
KarolCislowski, line 48: I can't send any new message to your API :( it returns an error 400 every time
image

KarolCislowski (changes requested):
Please fix your form 😺

KarolCislowski (approved):
Great work 🥇

PR #154, opened 2021-05-21
Happy thoughts Sara

Sartish:
API: https://saras-happy-thought.herokuapp.com
Frontend application: https://only-happy-thoughts.netlify.app/


server.js
```diff
@@ -14,15 +13,127 @@ mongoose.Promise = Promise
...
 const app = express()
 
+// specifying properties for the message, heart and createdAt
+// Date.now, creating date at the moment whenit is creating 
+
+const thoughtSchema = new mongoose.Schema({
+  message: {
+    type: String,
+    required: [true, "message is required"],
+    unique: true,
```
KarolCislowski, line 23: Why can't I be happy for the same thing that others are? I think the message is not a good candidate for a unique property.

server.js
```diff
@@ -14,15 +13,127 @@ mongoose.Promise = Promise
...
+    trim: true, 
+    minlength: 5, 
+    maxlength: 140
+    // validate: {
+    //   validator: (value) => {
+    //     return /^[^0-9]+$/.test(value);
+    //   }
+    // },
```
KarolCislowski, line 31: 🧹

server.js
```diff
@@ -14,15 +13,127 @@ mongoose.Promise = Promise
...
+        _id: id
+      },
+      { 
+        $inc: {
+          hearts: 1
+        }
+      },
+      {
+        new: true
+      }
```
KarolCislowski, line 91: 💯

server.js
```diff
@@ -14,15 +13,127 @@ mongoose.Promise = Promise
...
+//     if (updatedThought) {
+//       res.json(updatedThought)
+//     } else {
+//       res.status(404).json({ message: "not found" })
+//     }
+//   } catch {
+//     res.status(400).json({ message: "invalid request", error: "error" })
+//   }
+// })
+
```
KarolCislowski, line 136: remember to always restrict access to all endpoint which can manipulate your data stored in DB.

KarolCislowski (approved):
Nice API but IMHO you should remove unique property- if 3 other recruiters will try to send "testing" they will assume that it is broken 😭

PR #157, opened 2021-05-21
Happy Thoughts API - Maria Sparre 

MarySparrow:
Heroku: https://all-happy-thoughts.herokuapp.com/
Frontend: https://happy-thoughts-maria-sparre.netlify.app/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 .DS_Store
+.env
 .env.local
 .env.development.local
 .env.test.local
--- a/server.js
+++ b/server.js
@@ -54,7 +54,7 @@ app.get('/thoughts', async (req, res) => {
 app.post('/thoughts', async (req, res) => {
   try {
     const newThought = await new Thought ({
-      message: req.body}).save() 
+      message: req.body.message}).save() 
       res.status(201).json(newThought)
   } catch(error) {
     if (error.code === 11000) {
--- a/server.js
+++ b/server.js
@@ -17,7 +17,7 @@ const app = express()
 const thoughtSchema = new mongoose.Schema({
   message: {
     type: String,
-    required: true,
+    required: [true, 'Write something, not less than 5 characters'],
     unique: true,
     minlength: 5, 
     maxlength: 140,
@@ -63,15 +63,24 @@ app.post('/thoughts', async (req, res) => {
   res.status(400).json({ message: 'Could not send message', error})
   }
 })
-// app.post('/thoughts/:id/likes', async (req, res) => {
-//   const { id } = req.params
+app.post('/thoughts/:id/likes', async (req, res) => {
+  const { id } = req.params
 
-//   try{
-//     const updatedThought = await Thought.findByIdAndUpdate(id, { $inc: { hearts: 1 } })
-//   } catch (error) {
-   
-//   }
-// })
+  try {
+    const updatedThought = await Thought.findOneAndUpdate(
+      { _id: id},
+      { $inc: { hearts: 5 }},
+      { new: true}
+    )
+    if (updatedThought) {
+      res.json(updatedThought)
+    } else {
+      res.status(404).json({ message: 'Not found' })
+    }
+  } catch (error) {
+    res.status(400).json({ message: 'Invalid request', error })
+  }
+})
 //DELETE route
 app.delete('thoughts/:id', async (req, res) => {
   const { id } = req.params
@@ -102,7 +111,6 @@ app.patch('/thoughts/:id', async (req, res) => {
   }
 
 })
-
 
 
 // Start the server
--- a/server.js
+++ b/server.js
@@ -43,7 +43,7 @@ app.use(express.json())
 
 //GET route
 app.get('/', (req, res) => {
-  res.send(listEndpoints)
+  res.send(listEndpoints(app))
 })
 app.get('/thoughts', async (req, res) => {
   const thoughts = await Thought.find().sort({createdAt: 'desc'}).limit(20).exec()
--- a/server.js
+++ b/server.js
@@ -1,26 +1,59 @@
 import express from 'express'
-import bodyParser from 'body-parser'
 import cors from 'cors'
+import dotenv from 'dotenv'
 import mongoose from 'mongoose'
+import listEndpoints from 'express-list-endpoints'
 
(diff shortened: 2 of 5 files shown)


server.js
```diff
@@ -1,30 +1,124 @@
...
-//
-//   PORT=9000 npm start
-const port = process.env.PORT || 8080
+const port = process.env.PORT || 8081
 const app = express()
 
+const thoughtSchema = new mongoose.Schema({
+  message: {
+    type: String,
+    required: [true, 'Write something, not less than 5 characters'],
```
ThereseHag, line 20: This works perfectly well ✨ Maybe it would be a nice addition to also show something in the frontend, describing why it can't be posted, in case there are to few or to many chars though? Looks good in postman :)

server.js
```diff
@@ -1,30 +1,124 @@
...
 // Add middlewares to enable cors and json body parsing
 app.use(cors())
-app.use(bodyParser.json())
+app.use(express.json())
 
-// Start defining your routes here
+//GET route
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send(listEndpoints(app))
```
ThereseHag, line 47: Nice! It gives a great overview of all the endpoints with methods.

server.js
```diff
@@ -1,30 +1,124 @@
...
+})
+app.get('/thoughts', async (req, res) => {
+  // const { page, per_page } = req.query
+ 
+  const thoughts = await Thought.find().sort({createdAt: 'desc'}).limit(20).exec()
```
ThereseHag, line 52: Nice, short and clean code to sort and limit. Good job ✨

server.js
```diff
@@ -1,30 +1,124 @@
...
+  res.json(thoughts)
+})
+
+// POST route
+app.post('/thoughts', async (req, res) => {
+  try {
+    const newThought = await new Thought ({
+      message: req.body.message}).save()
```
ThereseHag, line 60: If you would like to make the code even shorter here, you could go with just await new Thought(req.body).save() here since it's only the message being included in the req body at the moment.

server.js
```diff
@@ -1,30 +1,124 @@
...
+  const { id } = req.params
+
+  try {
+    const deleteThought = await Thought.findOneAndDelete({ _id: id })
+    if (deleteThought) {
+  } else {
+    res.status(404).json({ message: 'Not found'})
+  }
+  } catch (error){
+    res.status(400).json({ message: 'Invalid request', error})
```
ThereseHag, line 98: Nice work with the try /catch- blocks here, and that you incude some information in the error messages. Good job ✨

server.js
```diff
@@ -1,30 +1,124 @@
...
+  }
+})
+// PATCH route
+app.patch('/thoughts/:id', async (req, res) => {
+  const { id } = req.params
+
+  try{
+    const updatedThought = await Thought.findByIdAndUpdate(id, { name: req.body.name })
```
ThereseHag, line 106: I think here you could use (id, req.body, {new: true }). I'm a bit unsure what the part " name: req.body.name "  will do here, since it's not included in the model? There is a message, but not a name at this point. Maybe you had one and removed it?

server.js
```diff
@@ -1,30 +1,124 @@
...
+    res.status(400).json({ message: 'Invalid request', error})
+  }
+
 })
 
+
 // Start the server
 app.listen(port, () => {
   // eslint-disable-next-line
   console.log(`Server running on http://localhost:${port}`)
```
ThereseHag, line 122: This console log could be removed before the final deploy :)

.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 .DS_Store
+.env
```
ThereseHag, line 3: Great that you remember to add the .env file here :)

ThereseHag (approved):
Great job with this project, you're a star 🌟 You're code is always clean and nice to go through and all your endpoints makes good sense. Just as an addition, I really like your front-end as well, and like how you have given it your own touch. 

Keep up the great work of yours! 🎉

server.js
```diff
@@ -1,30 +1,124 @@
...
-//   PORT=9000 npm start
-const port = process.env.PORT || 8080
+const port = process.env.PORT || 8081
 const app = express()
 
+const thoughtSchema = new mongoose.Schema({
+  message: {
+    type: String,
+    required: [true, 'Write something, not less than 5 characters'],
+    unique: true,
```
KarolCislowski, line 21: The message is not a good field to make it unique. if 3 recruiters come and send "testing" 2 of them will think it is not working 😭

server.js
```diff
@@ -1,30 +1,124 @@
...
+    trim: true,
+    minlength: 5, 
+    maxlength: 140,
+    // enum: ['']
+    //match: regex
```
KarolCislowski, line 26: 🧹

server.js
```diff
@@ -1,30 +1,124 @@
...
+  }
+})
+app.post('/thoughts/:id/likes', async (req, res) => {
+  const { id } = req.params
+
+  try {
+    const updatedThought = await Thought.findOneAndUpdate(
+      { _id: id},
+      { $inc: { hearts: 1 }},
+      { new: true}
```
KarolCislowski, line 76: 💯

server.js
```diff
@@ -1,30 +1,124 @@
...
+  } catch (error) {
+    res.status(400).json({ message: 'Invalid request', error })
+  }
+})
+//DELETE route
+app.delete('thoughts/:id', async (req, res) => {
+  const { id } = req.params
+
+  try {
+    const deleteThought = await Thought.findOneAndDelete({ _id: id })
```
KarolCislowski, line 92: If you are using _id field to manipulate your data there are a set of mongoose functions to do it faster. for example https://mongoosejs.com/docs/api.html#model_Model.findByIdAndDelete

server.js
```diff
@@ -1,30 +1,124 @@
...
+    if (updatedThought) {
+      res.json(updatedThought)
+    } else {
+      res.status(404).json({ message: 'Not found'})    
+    }
+  } catch (error) {
+    res.status(400).json({ message: 'Invalid request', error})
+  }
+
 })
```
KarolCislowski, line 116: remember to always restrict endpoints that modify data in your DB.

KarolCislowski (approved):
Nice API 🦄 Reconsider the unique property usage 😸

PR #190, opened 2021-05-24
Martin D. - Happy Thoughts API

Martno96:
A few hours past midnight due to Heroku build errors and admittedly poor planning, but better late than never!
One bug is known and that's when liking/hearting posts - the id:s sent don't match the ones in the database, for some reason. I haven't had time to investigate this properly, but I will fix this in the coming days when time allows.


server.js
```diff
@@ -1,30 +1,112 @@
...
+  try {
+    const updatedThought = await Thought.findByIdAndUpdate( id, { $inc: { hearts: 1} }, { new: true })
+    if (updatedThought) {
+      res.json(updatedThought)
+    } else {
+      res.status(404).json({ message: "Not found" })
+    }
+  } catch (error){
+    res.status(400).json({ message: "Invalid request", error})
+  }
```
MariaPogosjan, line 107: I noticed in the front end that your likes don't work. Maybe it is because you named it id rather than /thoughts/:thoughtId/like. I think Technigo mentioned in the brief to add it that way because of the frontend, so look how you fetch the likes in the frontend. Maybe you need to change the name there?

MariaPogosjan (approved):
Great job Martin! Very inspiring BE project, go go go! 💯 🥇

server.js
Martno96: Since :id and :thoughtId are slugs, their names are only for local reference, much like how a function's params can be named differently than the arguments passed into the function :) Regardless of name, they carry the id the same. But thanks for the suggestion anyways! ^^ It turned out the actual source of the bug was something as silly as the route's end being "likes" instead of "like"

docs.md
```diff
@@ -0,0 +1,16 @@
+Endpoints:
```
KarolCislowski, line 1: 👍🏻

server.js
```diff
@@ -1,30 +1,112 @@
...
+    type: String,
+    required: true,
+    validate: {
+      validator: (value) => {
+        return !/^\S{31,}$/.test(value)
+      },
+      message: "Message contains too long words. Please use max 30 characters per word."
+    },
+    minlength: 5,
+    maxlength: 140
```
KarolCislowski, line 23: Nicely configured field

server.js
```diff
@@ -1,30 +1,112 @@
...
+      { method: "POST", path: "/thoughts", description: "Posts a new thought. Min 4 characters, max 140. Words can't be longer than 30 characters." },
+      { method: "DELETE", path: "/thoughts/:id", description: "Deletes a thought with the _id provided as the slug." },
+      { method: "PUT", path: "/thoughts/:id", description: "Edits the message (only!) a thought with the _id provided as the slug." },
+      { method: "POST", path: "/thoughts/:id/likes", description: "The thought with the _id provided as the slug gets its likes/hearts increased by 1. Method is POST instead of PUT or PATCH because of the characteristics of the previously developed frontend." }
+    ]
+  })
+})
+
+app.get('/thoughts', async (req, res) => {
+  const allThoughts = await Thought.find().sort({ createdAt: -1 }).limit(20)
```
KarolCislowski, line 53: 👍🏻

server.js
```diff
@@ -1,30 +1,112 @@
...
+    const updatedThought = await Thought.findOneAndReplace({ _id: id }, { message: req.body.message }, { new: true })
+    if (updatedThought) {
+      res.json(updatedThought)
+    } else {
+      res.status(404).json({ message: "Not found" })
+    }
+  } catch (error){
+    res.status(400).json({ message: "Invalid request", error})
+  }
+})
```
KarolCislowski, line 94: Remember to restrict access to these endpoints because not all users should have access to these functionalities 🤓

server.js
```diff
@@ -1,30 +1,112 @@
...
+
+app.post('/thoughts/:id/like', async (req, res) => { //post request instead of patch as to fit the already developed frontend
+  const { id } = req.params
+  try {
+    const updatedThought = await Thought.findByIdAndUpdate( id, { $inc: { hearts: 1} }, { new: true })
```
KarolCislowski, line 99: 💯

KarolCislowski (approved):
Great API 👍🏻

PR #198, opened 2021-05-31
project19-Ati

takrami:
API: https://api-happy-thought.herokuapp.com/

LIVE: https://jovial-jackson-97a3e9.netlify.app/


backend/server.js
```diff
@@ -0,0 +1,154 @@
...
+
+const Thought = mongoose.model("Thought", thoughtSchema);
+
+app.get("/", (req, res) => {
+  res.send("Hello world");
+});
+
+app.get("/thoughts");
+app.get("/thoughts", async (req, res) => {
+  const thoughts = await Thought.find();
```
KarolCislowski, line 54: The messages supposed to be sorted from the newest.

KarolCislowski (changes requested):
Please repair the sorting of your thoughts.

backend/server.js
takrami: Hey :) Thanks for your comment. It's sorted now.

KarolCislowski (approved):
Great API 👍🏻

PR #209, opened 2022-01-06
Zancotti-happy-thoughts-api

Zancotti:


HedvigM (commented):
Well done Sabrina!
I have made some small comments, but overall is it well executed and you have as always made that little extra features that makes it stand out and be great!

PR #210, opened 2022-01-06
Project-Happy thoughts API - Jenny

jenquach:
https://project-happy-thoughts--api.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,8 +1,9 @@
 import express from 'express'
 import cors from 'cors'
 import mongoose from 'mongoose'
+import listEndpoints from 'express-list-endpoints'
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts"
+const mongoUrl = process.env.MONGO_URL || "mongodb+srv://jenquach:Koda2022@cluster0.d5xuu.mongodb.net/project-happy-thoughts--api?retryWrites=true&w=majority"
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -13,13 +14,84 @@ mongoose.Promise = Promise
...
+// Main endpoint
 app.get('/', (req, res) => {
-  res.send('Hello world')
+  res.send('Welcome to Happy thougths API')
+})
+
+// List of all the endpoints
+app.get('/endpoints', (req, res) => {
+  res.send(listEndpoints(app))
+})
```
DALA746, line 52: Nice that you created endpoints!

DALA746 (commented):
When I started to look at your project in frontend, the page was blank and no information was shown and there were errors in console regarding map is undefined. I cloned your frotend repo and had a look at your code. In your frontend in App.js on line 20 you need to change `.then((data) => setThoughts(data.response))` to `.then((data) => setThoughts(data))`. Because `data.response` gives undefined and `map()` doesn't work, but `data` gives an array and mapping is successful. Also it can be good to rename `fetchThoughts()` function to something else, like for example `getRequest()`. I was confused when I tried to read code, because you have fetch and inside fetch you have `fetchThoughts()`. Otherwise, your project is working. I can post messages and see all thoughts that you have in your database. Good job! :)

PR #214, opened 2022-01-07
w19 Elin Diczfalusy Happy thoughts api

Ajliin:


server.js
```diff
@@ -1,25 +1,101 @@
...
+
+app.post('/', async (req, res) => {
+  const { message } = req.body
+
+  // name: name is shorened to: name
+
+  try {
+    //Success
+    const newThought = await new Thought({
+      message: message,
```
katiewu1, line 67: can shorten `message: message` with just `message`

server.js
```diff
@@ -1,25 +1,101 @@
...
 app.use(cors())
 app.use(express.json())
 
 // Start defining your routes here
-app.get('/', (req, res) => {
-  res.send('Hello world')
+app.get('/', async (req, res) => {
+  try {
+    const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20)
+    res.json(thoughts)
```
katiewu1, line 53: Can throw in `status(200)` here

server.js
```diff
@@ -1,25 +1,101 @@
...
+      {
+        $inc: {
+          hearts: 1,
+        },
+      },
+      {
+        new: true, //updated document directly- find in documentary
+      },
+    )
+    console.log('req body', req.body)
```
katiewu1, line 94: can remove console log here :)

server.js
```diff
@@ -1,25 +1,101 @@
...
+    res.status(201).json({ response: updatedHeart, success: true })
```
katiewu1, line 95: maybe better with status code 200 here instead of 201

server.js
```diff
@@ -1,25 +1,101 @@
...
+  } catch (error) {
+    res.status(400).json({ response: 'No thought with that ID', sucess: false })
```
katiewu1, line 97: just a typo sucess -> success

katiewu1 (commented):
Hej Elin! Great work! :)
Hope you find my review useful. One bigger problem I came across is your env. variable. Unfortunately I could see your .env file, think you forgot to put it on gitignore.

PR #261, opened 2022-05-18
Happy-thoughts api

rawisou:
# Project Happy Thoughts API

Hello! Welcome to the happy-thoughts API. The data is stored in Mongo DB and is being pulled from there.

Get method: 
**/thoughts/** Users can use the following endpoints to see all thoughts

**/thoughts/:thoughtId** Users can view a specific thought using a specific _id.

Post method:
**/thoughts/** Users can use a post method in this endpoint to add more thoughts.
**/thoughts/:thoughtId/like** Users can use a post method in this endpoint to add likes.

When time allows, I may add an edit and a delete button in the frontend to hook with the delete and patch method in the backend.

## View the api

https://happy-thoughts-api-rawi.herokuapp.com/

## View the api in frontend 
https://illustrious-frangipane-e38a00.netlify.app/

### DIFF
--- a/routes/thoughts.js
+++ b/routes/thoughts.js
@@ -5,9 +5,13 @@ const router = express.Router()
 
 router.get('/', async (req, res) => {
     const thoughts = await Thought.find()
+    .sort({ createdAt: "desc" })
+    .limit(20)
     res.json(thoughts)
  })
 
+
+//I wanted to try out different syntax for promises (I could have used async and await here as well for consistency.)
 router.post('/', (req, res) => {
     const thought = new Thought({
         message: req.body.message,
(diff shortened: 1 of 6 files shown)


routes/thoughts.js
```diff
@@ -0,0 +1,78 @@
...
+
+router.get('/', async (req, res) => {
+    const thoughts = await Thought.find()
+    .sort({ createdAt: "desc" })
+    .limit(20)
+    res.json(thoughts)
+ })
+
+
+//I wanted to try out different syntax for promises (I could have used async and await here as well for consistency.)
```
Nosslexa, line 14: Nice to se another way of writing this.

Nosslexa (commented):
Great job with this weeks project! Your deployed site works great. your code is easy to read and understand. I like that you have tried another way than using the async await in your post router. i se no duplicates and its cool that you also tried with the update and delete.

PR #263, opened 2022-05-18
Week 19 HP-API Done

Dopest0727:


server.js
```diff
@@ -1,27 +1,96 @@
...
+// ROUTES
+app.get('/', (req, res) => {
+  res.send('Tjaba, this is my API for my frontend app, see it here "https://tiopunkter.netlify.app"')
+})
+
+// ENDPOIJNT SHOWS 20 LATEST MSG's
+app.get('/thoughts', async (req, res) => {
+  const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20).exec()
+  res.status(200).json(thoughts)
+})
```
Laura-Sjolander, line 51: I tried this endpoint in backend  Seems to work fine, All the messages/thoughts is there.

server.js
```diff
@@ -1,27 +1,96 @@
...
+const Thought = mongoose.model('Thought', ThoughtSchema)
+
+// ADD MIDDLEWARE
+app.use(cors())
+app.use(express.json())
+
+// ROUTES
+app.get('/', (req, res) => {
+  res.send('Tjaba, this is my API for my frontend app, see it here "https://tiopunkter.netlify.app"')
+})
```
Laura-Sjolander, line 45: Tried this route, works well! Great!

server.js
```diff
@@ -1,27 +1,96 @@
...
+  const updatedHearts = await Thought.findByIdAndUpdate(
+    thoughtId, 
+    { 
+      $inc: { 
+        hearts: 1
+      }
+    },
+    {
+      new: true
+    }
```
Laura-Sjolander, line 78: Nice way to update the hearts by incrementing

Laura-Sjolander (commented):
Nice and clean code! Works well, I tried both the backend and the frontend, connection between them is there and stores messages as it should. 
And I love the design in the frontend part, as always. 
Keep on doing what you are doing!

server.js
```diff
@@ -1,27 +1,96 @@
...
+  )
+
+  if (updatedHearts) {
+    res.json(updatedHearts)
+  } else {
+    res.status(404).json({ response: 'No happy thought with this ID', success: false })
```
jessand77, line 84: Good that you used the 404 status code here.

server.js
```diff
@@ -1,27 +1,96 @@
...
+app.use(cors())
+app.use(express.json())
+
+// ROUTES
+app.get('/', (req, res) => {
+  res.send('Tjaba, this is my API for my frontend app, see it here "https://tiopunkter.netlify.app"')
+})
+
+// ENDPOIJNT SHOWS 20 LATEST MSG's
+app.get('/thoughts', async (req, res) => {
```
jessand77, line 48: A suggestion is to use a try and catch block here even if there is not so much that should be able to go wrong here :-)

jessand77 (commented):
Good job with this week's project too :-) Everything seems to work as it should, the code has a good structure and includes error messages and validation.

PR #266, opened 2022-05-19
Happy thoughts api - Vanessa

vanhaj:
https://happy-thoughts-projectapi.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -13,7 +13,6 @@ const Thought = mongoose.model('Thought', {
     required: true,
     minlength: 5,
     maxlength: 140,
-    trim: true,
   },
   like: {
     type: Number,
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,11 +1,39 @@
...
+    type: Date,
+    default: () => new Date(),
+  },
+  // createdAt: {
+  //   type: String,
+  //   default: () =>
+  //     new Date(Date.now()).toLocaleString('sv-SE', {
+  //       timeZone: 'Europe/Stockholm',
+  //     }),
+  // },
```
Thereese, line 32: I think you already know this, but your time is still showing 2 hours behind in the database. I compared my code with yours, because I´ve used the one that you commented out, checked both backend and frontend and they are the same. Maybe it was something else that crashed it?

server.js
```diff
@@ -17,11 +45,60 @@ app.use(cors());
...
+});
+
+app.post('/happy-thoughts/:thoughtId/like', async (req, res) => {
+  const { thoughtId } = req.params;
+  try {
+    const likeToUpdate = await Thought.findByIdAndUpdate(thoughtId, {
+      $inc: { like: 1 },
+    });
+    res.status(200).json({
+      response: `Like ${likeToUpdate.message} has been updated`,
```
Thereese, line 91: I like that you´ve added error-messages, makes the error a bit nicer:)

server.js
```diff
@@ -17,11 +45,60 @@ app.use(cors());
...
+    res.status(400).json({ response: error, success: false });
+  }
 });
 
 // Start the server
 app.listen(port, () => {
   console.log(`Server running on http://localhost:${port}`);
 });
+
+// https://happy-thoughts-projectapi.herokuapp.com/
```
Thereese, line 104: Maybe remove if it´s not used?

Thereese (approved):
Great job, VanHaj! Works like a charm and extra points for connecting it successfully to frontend!🤩

PR #267, opened 2022-05-19
WK 19 Happy thoughts API

MarianneArdin:


server.js
```diff
@@ -16,12 +17,92 @@ const app = express();
...
+		type: Date,
+		default: () => new Date(),
+	},
+});
+
+const thought = mongoose.model('thought', ThoughtsSchema);
+
+//Start of routes
+app.get('/', (req, res) => {
+	const Main = {
```
LovisaBrorson, line 42: This const Main was new to me, clever to put a little longer message in there!

PR #269, opened 2022-05-19
Happy Thoughts API

rijad90:


README.md
```diff
@@ -1,13 +1,3 @@
-# Project Happy Thoughts API
-
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
-
```
EmmaaBerg, line 6: You could add some text about the project, the purpose for example so if someone is viewing your project they can quickly get a sense of what it's about.

EmmaaBerg (commented):
Good job with the Happy Thoughts API! It works fine with your frontend and reaches the requirements. I read the common pitfalls about writing good git commits, I myself need to practice more on that as well. But maybe you could try to do more commits for the next project and add more descriptive messages. I guess it will be important in future jobs when we will work in teams on a project.

PR #273, opened 2022-05-19
Savannah - Happy Thoughts MERN Stack

savannah-hayes:
Frontend: https://happy-thoughts-frontend.netlify.app/

Backend: https://happy-thoughts-mern.herokuapp.com/


client/src/components/ThoughtForm.js
```diff
@@ -0,0 +1,66 @@
...
+        type="text"
+        id="thoughtAuthor"
+        className="author-input"
+        value={username}
+        onChange={handleUser}
+        placeholder="Enter your name or post anonymously..."
+      />
+      {error &&
+        <ul className="error-message">
+          {errorMessages.map(message => (
```
jenniferwvng, line 42: Nice implementation

client/src/components/partials/TimeStamp.js
```diff
@@ -0,0 +1,57 @@
...
+  const timeStamp = () => {
+    const seconds = Math.floor((new Date() - new Date(props.createdAt).getTime()) / 1000);
+    let interval = Math.floor(seconds / 31536000);
+    let unitsOfTime;
+
+    if (interval >= 1) {
+      unitsOfTime = "year";
+    } else {
+      interval = Math.floor(seconds / 2592000);
+      if (interval >= 1) {
```
jenniferwvng, line 13: Hmm at first glance I'm not sure what this does since it's using the same comparator for every if-statements, but I'm sure I'm missing something since you put it there for a reason, so commenting just because I'm curious about how it works

server/config/db.js
```diff
@@ -0,0 +1,14 @@
+import mongoose from "mongoose";
+
+const mongoUrl = process.env.CONNECTION_URL || "mongodb://localhost/thoughts";
+
+const connectDB = () => {
+  try {
+    mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
+  } catch (error) {
+    console.error(`Error: ${error.message}`)
+    process.exit(1)
```
jenniferwvng, line 10: Interesting, haven't seen this yet so learned something new to google

jenniferwvng (commented):
Clear code and nice structuring of the files, interesting to follow the code

client/src/components/partials/TimeStamp.js
savannah-hayes: The first thing I do is calculate the difference of the time it is now compared to the time a happy thought was created on in milliseconds and divide by 1000 to get that time in seconds. I store that time in a variable called **seconds** on line 5.

Then the variable **interval** which is created on line 6 converts time intervals given in seconds into more human readable form. 

Maybe this breakdown will help you understand it better:
`secondsInAYear = 31536000`
`secondsInAMonth=2592000`
`secondsInADay=86400`
`secondsInAnHour=3600`
`secondsInAMinute=60`

Basically, what is happening is that I take the time the happy thought was created on in seconds and divide it by how many seconds there are in a year, month, day, hour, and minute.

And if the interval is greater than or equal to one I want to add the units of time in string format ("year", "month", "day", "minute", "second") to the end of the actual time that happy thought was created on which is the variable **interval**.

savannah-hayes: I hope that explains how it works in a clear way.

PR #274, opened 2022-05-20
Week 19 - Sofie Pellegrini

SofiePellegrini:


server.js
```diff
@@ -16,9 +17,87 @@ const app = express();
...
+    'Hello!': "This is Sofie's Happy Thoughts API.",
+    'Here you can watch the API live':
+      'https://graceful-lily-1fde84.netlify.app/',
+    'And see all endpoints here': '/endpoints',
+  });
+});
+
+app.get('/endpoints', (req, res) => {
+  res.send(allEndpoints(app));
+});
```
rijad90, line 52: nice route explaining what this is and also that you added the allEndpoints, makes my job easier when reviewing :)

rijad90 (approved):
I tested both your frontend and backend, both are working as expected and I dont have nothing else to say than Good Job Sofie! Your backend skills are really good :)

PR #280, opened 2022-05-20
Project happy thoughts api week 19

Mimmisen:
This project deployed link: https://project-happy-thoughts-week-19.herokuapp.com/

link to the frontend connected to this API: https://happy-thoughts-mimmi.netlify.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -3,8 +3,6 @@ import cors from "cors";
 import mongoose from "mongoose";
 import { type } from "express/lib/response";
 
-// import listEndpoints from "express-list-endpoints";
-
 const mongoUrl =
   process.env.MONGO_URL || "mongodb://localhost/happyThoughts-api";
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -15,10 +17,81 @@ const app = express();
...
+  res.send(listEndpoints(app));
+});
+
+app.get("/thoughts", async (req, res) => {
+  try {
+    const thoughts = await Thought.find({})
+      .sort({ createdAt: "desc" })
+      .limit(20)
+      .exec();
+    res.status(201).json(thoughts);
```
tiiliu, line 60: This is a small and totally optional thing, but if you would like to have your code be as symmetrical as possible in the backend, you could implement the response here in similar style as you do in the error response, having an object with two properties, one for success: true and the other showing the data,  in a way you want to name it (response: thoughts/data: thoughts). I also noticed that you have used this in other paths, with response so maybe it will also be more unified to use the same approach throughout the code 😊

tiiliu (commented):
Great work Mimmi! Your code fulfills all the requirements and works without any problems. You have good and consistent validation in your code for errors that works and your happy thoughts page is so cute 😍 🎊

PR #288, opened 2022-05-21
Week 19- Happy thoughts Api- Sara Boström Bundock

bostrombundock:


EmmaSprings (commented):
Well done with your API ang mongoDB this week! I can send happy thoughts without an issue :) I couldn't send a like though, having issues with that myself, so will have to investigate further on that. 
 Only thought I had was that it would be nice to see the endpoints when you go to the initial page of the API (even if I know what it is :) ) Perhaps install express list endpoints, then it does that for you. Easy! :)

server.js
```diff
@@ -1,24 +1,96 @@
...
+
+    });
+ }
+});
+
+ app.post('/thoughts/:thoughtId/like', async (req, res) => {
+   const { thoughtId } = req.params;
+
+try {
+  const newMessage = await new thought.findByIdAndUpdate(thoughtId, {$inc: {hearts: 1}})
```
savannah-hayes, line 82: I still noticed you couldn't add likes and after testing out your code and really inspecting it. It was because you some how typed `await  new thought.findByIdAndUpdate` but you don't need to use the **new** keyword here which was throwing off your API. 

so if you change line 82 to this it should fix your post route to add likes:
`const newMessage = await thought.findByIdAndUpdate(thoughtId, {$inc: {hearts: 1}})`

server.js
```diff
@@ -1,24 +1,96 @@
...
+  res.status(200).json(newMessage);
+   
+} catch(error) {
+  res.status(400).json({
+    message: 'Please try again!',
+    error: error.errors,
```
savannah-hayes, line 88: I also noticed while testing this route to try to get the likes to add that this `error: error.errors` wasn't actually outputting anything in postman when an error occurred. 

So if you would want to send a message with more details about the error you could use this:
`error: error.message` this will send a nice error message to the frontend and give them a better understanding of what errors might have occurred.

server.js
```diff
@@ -1,24 +1,96 @@
...
 
-// Add middlewares to enable cors and json body parsing
+
 app.use(cors());
 app.use(express.json());
 
+const ThoughtSchema = new mongoose.Schema({
+  message: {
+    type: String,
+    required: true,
```
savannah-hayes, line 22: Nice job adding required here and setting it equal to true!

server.js
```diff
@@ -1,24 +1,96 @@
...
+    });
+  }
+});
+      
+app.post('/thoughts', async (req, res) => {
+  const { message } = req.body;
+
+  try {
+    const newThought = await new thought({message}).save();
+    res.status(200).json(newThought)
```
savannah-hayes, line 66: Did you know you could use a status code of 201 here ie `res.status(201)` ? In this case a status 201 would give the user a clearer status code, meaning that a new happy thought was successfully created.

server.js
```diff
@@ -1,24 +1,96 @@
...
+app.get('/thoughts', async (req, res) => {
+  try {
+  const thoughts = await thought.find().sort({createdAt:'desc'}).limit(20).exec();
+  res.status(200).json(thoughts);
+
+} catch(error) {
+    res.status(400).json({
+    message:'Please try again',
+    error: error.errors,
+    });
```
savannah-hayes, line 57: Great job implementing try catch blocks in your code!

savannah-hayes (commented):
Great job on your Happy Thoughts Api! I only found a small bug in your code and made a comment about it in your server.js file. It seems to just be a typo that can be easily fixed. 

I checked both the frontend and backend code and deployed sites and they look good. You did a great job and seem to have met all of the requirements for this week. Nice work :)

PR #291, opened 2022-05-22
Project Happy Thoughts / Text your Ex API - Michael

michaelchangdk:


server.js
```diff
@@ -16,9 +16,155 @@ const app = express();
 app.use(cors());
 app.use(express.json());
 
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next();
+  } else {
+    res.status(503).json({ error: "Service unavailable" });
+  }
+});
```
mathildakarlsson, line 25: Cool that you added status code 503!

server.js
```diff
@@ -16,9 +16,155 @@ const app = express();
...
+
+const TextSchema = new mongoose.Schema({
+  message: {
+    type: String,
+    required: [true, "Please type a message."],
```
mathildakarlsson, line 30: I like the error information you added to the user in the schema!

server.js
```diff
@@ -16,9 +16,155 @@ const app = express();
...
+  },
+  likes: {
+    type: Number,
+    default: 0,
+  },
+  createdAt: {
+    type: Date,
+    default: () => new Date(),
+  },
+  ex: {
```
mathildakarlsson, line 49: Great job, this is a red level goal to add a "username"!

server.js
```diff
@@ -16,9 +16,155 @@ const app = express();
...
+});
+
+const Text = mongoose.model("Message", TextSchema);
+
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send([
+    { "/": "Hi! This is the text your ex backend" },
+    { "https://textyourex.netlify.app/": "Frontend here!" },
```
mathildakarlsson, line 70: Very smart to add frontend link in this endpoint/landingpage of the api

server.js
```diff
@@ -16,9 +16,155 @@ const app = express();
...
+  ]);
+});
+
+// GET ALL TEXTS - LIMIT AT 20
+app.get("/texts", async (req, res) => {
+  try {
+    const texts = await Text.find()
+      .sort({ createdAt: "desc" })
+      .limit(40)
```
mathildakarlsson, line 79: Hmm, should it be `.limit(20)` here instead of 40?

server.js
```diff
@@ -16,9 +16,155 @@ const app = express();
...
+    res.status(400).json({
+      success: false,
+      status_code: 400,
+      message: "Bad request, could not fetch thoughts.",
+      error: err.errors,
+    });
+  }
+});
+
+// POST A NEW TEXT
```
mathildakarlsson, line 92: Good structuring with your headline comments

server.js
```diff
@@ -16,9 +16,155 @@ const app = express();
...
+    res.status(200).json({ success: true, data: texts });
+  } catch (err) {
+    res.status(400).json({
+      success: false,
+      status_code: 400,
+      message: "Bad request, could not fetch thoughts.",
+      error: err.errors,
+    });
+  }
+});
```
mathildakarlsson, line 142: Cool! Here's a black level req!

server.js
```diff
@@ -16,9 +16,155 @@ const app = express();
...
+
+// SEARCH FOR EX TEXTS BY NAME - LIMIT TO 20 TEXTS - THIS ISN'T WORKING
```
mathildakarlsson, line 144: Too bad it isn't working, it would be a nice feature. But you'll definitely be able to implement it later on if you like!

mathildakarlsson (commented):
Hi Michael!

Great job with your API! I am impressed by your implementation of the stretch goals. And a huge bonus is of course your front end and styling. The theme of the application is perfect 😂  

Your code is as always super well written and I can't complain about anything really! I saw that you commented that you want to limit the texts to 20 but in your code you've written .limit(40) on line 79. But I see that in your sorting you've put the limit to 20, so maybe the (40) has to do with that somehow? 

Other than that I don't really find anything to improve so I just want to say well done with both back end and front end!

PR #295, opened 2022-05-22
 project-happy-thoughts-api

Kras053:
https://send-happy-thought-api.herokuapp.com/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,26 @@
 # Project Happy Thoughts API
 
-Replace this readme with your own information about your project.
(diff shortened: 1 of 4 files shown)


SteppbySteph (commented):
Even though there weren't a lot of code that had to be written for us during this week it sure took a long time finishing it. Well done on this project! All three endpoints were tested out on Postman and they work as they should giving us the expected outcome from the project brief. Thank you for writing such a good content in your README file. Good luck on your future adventure with backend!

PR #296, opened 2022-05-22
Happy thoughts API - Emma Springs

EmmaSprings:


server.js
```diff
@@ -16,11 +19,107 @@ const app = express();
...
+   .sort({ 
+    createdAt: 'desc'
+  })
+  .limit(20)
+  .exec()
+
+  res.status(200).json(thoughts)
+
+} catch (error) {
+  res.status(400).json({success: false, response: error})
```
sofiaringstedt, line 63: A tip here is that if you write error: err.errors, you will get access to the built in error message in the js error object!  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

server.js
```diff
@@ -16,11 +19,107 @@ const app = express();
...
+}
+})
+
+//Post a new thought
+
+app.post('/thoughts', async (req, res) => {
+
+const { message } = req.body
+
+const thought = await new Thought({ message: message}).save()
```
sofiaringstedt, line 73: Just realized that I declared and saved the thought variable inside the try() function, I wonder what difference it makes in the outcome!

sofiaringstedt (commented):
Well done this week Emma! As you mentioned yourself in the readme, you have this loading error in your like function. I had one suggestion that I added in the review, about await. Otherwise, do you remember if you had this error prior to adding your own api? 

Also so nice to see your very nice looking frontend! 🥇

PR #299, opened 2022-05-22
Week-19 project-happy-thoughts-api

toal13:
https://week-19-happythoughts.herokuapp.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,27 +1,101 @@
-import express from "express";
-import cors from "cors";
-import mongoose from "mongoose";
+import express from "express"
+import cors from "cors"
+import mongoose from "mongoose"
+import listEndpoints from "express-list-endpoints"
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/happyThoughts";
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -1,27 +1,101 @@
...
+    const updatedHearts = await Thought.findByIdAndUpdate(
+      thoughtId,
+      {
+        $inc: {
+          hearts: 1,
+        },
+      },
+      {
+        new: true,
+      }
```
amandatange, line 84: Im curious what this property does and where it's used? I must've missed that if it was in one of the videos? Is it something for the frontend perhaps, would be interested to know more! :)

server.js
```diff
@@ -1,27 +1,101 @@
...
-});
+app.get("/", (req, res) =>
+  res.send(listEndpoints(app)))
+
+app.get("/thoughts", async (req, res) => {
+  const allThoughts = await Thought.find()
+    .sort({ createdAt: "desc" })
+    .limit(20)
+    .exec()
+  res.json(allThoughts)
```
amandatange, line 51: nicely done!

amandatange (commented):
Sorry for the lack of comments, there wasn't that much to comment on because I think it is done really well and works great! I would have liked to see a frontend connected to your api, but that should probably not be an issue connecting them since it looks like your backend works great :D

PR #318, opened 2022-12-01
Week 15 Kaja Wilbik 

KajaWilbik:


RosenholmJohanna (commented):
I tried the API with POST and GET using postman and it seems to work fine! Clean code, easy to follow and I also like that  you list the endpoints to use. God job :)

PR #330, opened 2022-12-02
Project-happy-thoughts-api  by Tina Bruce

brucette:
Link to deployed project:
https://project-happy-thoughts-api-d6aenh5q2a-lz.a.run.app/

Frontend:
https://happy-thoughts.brucette.com/

Frontend repo: 
https://github.com/brucette/project-happy-thoughts

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,3 +25,6 @@ https://project-happy-thoughts-api-d6aenh5q2a-lz.a.run.app/
 View it when used in a frontend:
 https://happy-thoughts.brucette.com/
 
+Frontend repo: 
+https://github.com/brucette/project-happy-thoughts
+
--- a/README.md
+++ b/README.md
@@ -1,13 +1,27 @@
 # Project Happy Thoughts API
 
-Replace this readme with your own information about your project.
+The aim of the project was to build an API that includes handling of user input and works well with an existing frontend, and to:
 
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+✓ use POST requests to send data to the API
+
+✓ store data in a database from those POST requests
+
+✓ validate data and ensure the database only contains 'good' data
 
 ## The problem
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -2,26 +2,90 @@ import express from "express";
...
+  } else {
+    res.status(503).json({error: 'Service unavailable'});
+  }
+});
+
+/// Lists the enpoints available:
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.json(listEndpoints(app));
+});
```
jesshansson, line 54: If you have time, you could also create a json-message where you print out the endpoints and what you can see/do on them. Easier to understand for the broad mass :) And also you could add the adress to your frontend page there!

README.md
```diff
@@ -1,13 +1,30 @@
...
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+MongoDB was used as the database and Mongoose to model and store the data. For the API the following endpoints were made to fit the existing frontend project "Happy Thoughts":
+
+**/thoughts** - get 20 of the latest thoughts currently stored in the database using GET **OR** POST and new thought to be added to the database provided it passes validation 
+
+**/thoughts/:id/like** - increase an existing thoughts like count (for when user clicks the heart button on the frontend)
+
```
jesshansson, line 18: This you could also add on the first page in your API :)

jesshansson (commented):
Great job! You code is really clean and easy to follow. You've also added good comments that explains what the code does. Not much to complain about at all 😄 

The thing you could look up though is the middleware that you use now (see my comment), as I think it could be the thing that now makes the error message sometimes show instead of your API, even if your db is connected. A slow connection. either from Google cloud or from my wi-fi, kind of messes that up, I've come to discover myself! (I can see your API after a while though, so everything works 👍 )

Looking at your frontend page, something seem to happen when I press the like-button, the message dissapear I think? So extremely sorry if I emptied your API! 😰 Can't like the new ones either 🧡

PR #337, opened 2022-12-04
Linda Malms happy-project

malmen237:
View the backend live
https://project-happy-thoughts-api-auhzlcxnrq-lz.a.run.app/

View the frontend live
https://lindas-happy-thoughts.netlify.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -21,6 +21,14 @@ app.use((req, res, next) => {
   }
 });
 
+mongoose.set('toJSON', {
+  virtuals: true,
+  versionKey: false,
+  transform: (doc, converted) => {
+    delete converted._id;
+  }
+});
+
 // MongoDB
 const MessageSchema = new mongoose.Schema({
   message: {
--- a/server.js
+++ b/server.js
@@ -6,9 +6,6 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
 mongoose.Promise = Promise;
 
-// Defines the port the app will run on. Defaults to 8080, but can be overridden
-// when starting the server. Example command to overwrite PORT env variable value:
-// PORT=9000 npm start
 const port = process.env.PORT || 8080;
 const app = express();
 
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -6,22 +6,100 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
...
 app.use(cors());
 app.use(express.json());
 
+app.use((req, res, next) => {
+  if (mongoose.connection.readyState === 1) {
+    next()
+  } else {
+    res.status(503).json({ error: "Service unavaliable" })
+  }
+});
```
brucette, line 22: This code checking the connection to the database might be causing "service unavailabe" to be returned when the API is too slow in returning a response, even if the db is connected, and requiring a refresh before the API data is shown.

server.js
```diff
@@ -6,22 +6,100 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
...
+
+mongoose.set('toJSON', {
+  virtuals: true,
+  versionKey: false,
+  transform: (doc, converted) => {
+    delete converted._id;
+  }
+});
```
brucette, line 30: I don't fully understand this code but assume it is for removing the _ from the id. Nice perseverance on getting that done! 😎

server.js
```diff
@@ -6,22 +6,100 @@ const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
...
+    default: () => new Date()
+  }
+})
+
+const Message = mongoose.model("Message", MessageSchema);
+
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send([
```
brucette, line 54: Nice, clear presentation of the endpoints.

brucette (commented):
Great job this week Linda! I tried really hard to find something to poke holes at but the only thing I can really come up with is to have some more comments in the code :D 

Otherwise clean, structured code and the site works well. I like the separate path and url you have on the starting page and with the middle wear removed, makes it easy to use and without clutter.

I think it might be because of google cloud being slow that when I initially clicked on your backend link, I got the error message “service unavailable”. After refreshing the page it returned everything fine. I noticed the same issue with my page also. 


Just a side note on the frontend - The background colour for liked buttons isn’t different from the non-clicked ones upon loading the page, only when clicked do they turn pink - but when the page is refreshed the buttons background colour changes back to grey (so they are all the same colour whether like count is 0 or not).

PR #344, opened 2022-12-04
project-happy-thoughts-api, Neah Rockstroh

Archkrull:


server.js
```diff
@@ -17,10 +18,69 @@ app.use(cors());
...
 // Start defining your routes here
+
+////////////////
+const ThoughtsSchema = new mongoose.Schema({
+  message: {
+    type: String,
+    required: true,
+    minlength: 1,
+    maxlength: 140,
+    trim: true
```
Sofia-Andersson, line 29: trim:true, I guess this makes the message 140 letters if it's too long. Good idea! :)

server.js
```diff
@@ -17,10 +18,69 @@ app.use(cors());
...
+   console.log(req.body);
+   try {
+    const newThought = await new Thoughts({message: message, createdAt: createdAt}).save();
+    res.status(201).json({success: true, response: newThought});
+   } catch(error) {
+     res.status(400).json({success: false, response: 'cannot post mesage'});
+   }
+ });
+
+app.patch("/thoughts/:id/hearts", async (req, res) => {
```
Sofia-Andersson, line 74: Good that you tried to use patch which is (I think) new to us! 👍

Sofia-Andersson (commented):
Well done last week Neah! 😊⭐️

- Am I able to understand the code easily? Does the code have lots of duplicates? Could it be tidied up?
I'm able to understand the code easily! Some comments could be added if you would like to make it even more easy to understand each part of the code.

- Does it work or does anything look broken?
Seems to work fine to me.

- Does it follow the general and blue requirements?
Yes!

Good work Neah, as mentioned some comments and more explanations in the readme could make lite easier for anyone else reading the code. 🤔

PR #347, opened 2022-12-04
Happy thoughts API 

marwebdesign:


marwebdesign:
Google cloud: 
https://project-happy-thoughts-api-ebt3b63p3a-lz.a.run.app/thoughts

Frontend: 
https://github.com/marwebdesign/Project-Happy-Thoughts-API-Frontend

Link to Netlify: 
https://silver-pasca-18ad6f.netlify.app/

server.js
```diff
@@ -18,9 +18,71 @@ app.use(express.json());
...
+app.get("/thoughts", async (req, res) => {
+  const response = {
+    success: true,
+    body: {}
+  }
+  try {
+    response.body = await Thoughts.find().sort({ createdAt: "desc" }).limit(20).exec();
+    res.status(200).json({
+      success: true,
+      body: response
```
LinneaAjger, line 64: Why did you choose to put the "response" variable inside the body? When I look at your json file, it's a double success & body nested inside.

server.js
```diff
@@ -18,9 +18,71 @@ app.use(express.json());
...
+  }
+})
+
+app.patch("/thoughts/:thoughtId/like", async (req, res) => {
+  const { thoughtId } = req.params
+  try {
+    const thoughtToAddLike = await Thoughts.findByIdAndUpdate(thoughtId, { $inc: { hearts: 1 } })
+    res.status(200).json({ success: true, response: `One more like for ${thoughtToAddLike.message}` })
+  } catch (error) {
+    res.status(400).json({ success: false, response: error })
```
LinneaAjger, line 82: Here you use "response" instead of "body" as above. I don't know whats correct but I guess it's good to be consistent. If there's a reason I would love to know!

LinneaAjger (commented):
Good job Maria! Your Api works great and makes it possible to post and get thoughts 😊 But when I posted a super long one to test the validation and it apparently got through 🙇‍♀️ The short one (<4) also, so that would be super good if you could fix (I had the same issue so holla at me if you get stuck)! And I'm sorry for messing up your page with that  long one 🙈 
Otherwise than that you meet all the requirements! 🌟 Go Tigers! // Linnéa

PR #356, opened 2022-12-04
HappyThoughts- Agnes

AgnesAhlman:


karlfredrikholm:
Nice that you put the mongoose schema and model in another file, made server.js even more readable!  Had to look in the front-end repo as well, and really liked the refactoring with all the api-business in one separate file. 🫵🤩

PR #359, opened 2022-12-06
Kristin Larsson- Project Happy Thoughts API

Kristin-Larsson:
https://project-happy-thoughts-api-ltlv2yd2lq-lz.a.run.app

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,8 +1,8 @@
-import express from "express";
+import express, { response } from "express";
 import cors from "cors";
 import mongoose from "mongoose";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-happythought";
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
 mongoose.Promise = Promise;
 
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -18,7 +18,79 @@ app.use(express.json());
...
+  } catch (err) {
+    res.status(400).json({
+      message: "Could not find this post",
+      response: err.errors,
+      success: false,
+    });
+  }
+});
+
+app.patch("/members/:id/like", async (req, res) => {
```
charlottejohansson, line 86: Don't forget to double check the id-name in your happy thought project when you link together the project :)

server.js
```diff
@@ -18,7 +18,79 @@ app.use(express.json());
...
+   const { id } = req.params;
+   try {
+    const thoughtToUpdate = await Thought.findByIdAndUpdate(id, {$inc: {heart: 1}});
```
charlottejohansson, line 89: Same with name of the "heart", double check the name in your happy thought project - otherwise it maybe does not work even if the code is correct :)

server.js
```diff
@@ -18,7 +18,79 @@ app.use(express.json());
...
+    response: thoughts
+    })
+  } catch (err) {
+    res.status(400).json({
+      message: "Could not find this post",
+      response: err.errors,
+      success: false,
+    });
+  }
+});
```
charlottejohansson, line 84: Looks good and easy to follow.

charlottejohansson (commented):
Good job Kristin - you have some more work to reach the requirement but otherwise your code is easy to follow :)

PR #368, opened 2023-05-17
https://project-happy-thoughts-api-sjzdfi57eq-lz.a.run.app

EmiliaGA:
https://project-happy-thoughts-api-sjzdfi57eq-lz.a.run.app

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "@babel/preset-env": "^7.16.11",
         "cors": "^2.8.5",
         "express": "^4.17.3",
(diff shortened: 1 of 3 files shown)


cecavi (commented):
Nice job on your server.js. I see you matched the endpoints perfectly for our former project happy thoughts. I saw in your happy-thoughts-project repo that you placed your deployed API in the fetch. If I´m not wrong it seems we share a similar problem with the heart likes not updating. Maybe we can help each other out and solve it.

But all in all a very good job!

PR #382, opened 2023-05-19
Happy Thoughts API

FionaKlacar:


server.js
```diff
@@ -15,13 +37,77 @@ const app = express();
...
+      response: err,
+      message: "Error, no thoughts found"
+  })
+}
+})
+
+// POST - post a new thought 
+app.post('/thoughts', async (req, res) => {
+  const { message } = req.body; // removed createdAt so that the user cannot input the date, and it will always set to default
+  const thoughts = new Thought({ message, hearts: 0 }); // hearts always 0, but can still update using patch request
```
beemailley, line 69: You don't need to include "hearts: 0" here. You've already set the default in the Schema to 0, so it will be set to zero when a new thought is posted.

server.js
```diff
@@ -15,13 +37,77 @@ const app = express();
...
+  try {
+    const savedThought = await thoughts.save();
+    res.status(201).json(savedThought);
+  }
+  catch(err) {
+    res.status(400).json({
+      message: "Could not save thought to the database", error: err.errors
+    })
+  };
+})
```
beemailley, line 79: In your GET endpoint, you have a more comprehensive response for both the successful attempt and the failed attempt (success, response, and message). I think Daniel has said that it's good practice to have the same response structure for all endpoints. So, the POST (and PATCH) should have the same response structures.

server.js
```diff
@@ -15,13 +37,77 @@ const app = express();
...
+
+// PATCH thoughts/:thoughtId/like - get a heart like on a thought 
+app.patch('/thoughts/:thoughtId/like', async (req, res) => {
+  const { thoughtId } = req.params;
+  try {
+    const thought = await Thought.findById(thoughtId);
+    if (thought) {
+      thought.hearts += 1;
```
beemailley, line 87: Do you know if using 
`thought.hearts =+ 1; `
acts any differently than
`thought.hearts + 1; ` 
with no "="?

beemailley (commented):
Overall it looks good!  Sounds like you did a good job of troubleshooting some issues, and it all works right!

PR #383, opened 2023-05-19
Happy-thoughts-API

Yu-Miao222:
front-end : https://happy-thought-by-yumiao.netlify.app/
back-end : https://project-happy-thoughts-api-s7fgmoqnaa-lz.a.run.app

### DIFF
--- a/server.js
+++ b/server.js
@@ -91,9 +91,9 @@ app.post("/thoughts", async(req,res)=>{
 })
 //Endpoint 3: POST/thoughts/:thoughtId/like, like a post by inreasing the hearts by 1
 app.post("/thoughts/:thoughtId/like", async(req,res) => {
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -18,9 +18,121 @@ app.use(express.json());
...
+
+  } catch (e) {
+    res.status(400).json({
+      success: false,
+      response: e,
+      message: "can not post your thought"
+    });
+  }
+
+})
```
khleoth, line 91: The update function is very similar but could have been good to implement for practice purposes, but this way works fine aswell!

khleoth (commented):
Well done Yu, all seems to be working as intended! Did not see any delete function on the front-end but it's here on the backend, although it's always good to bring along for practice in the back-end. Keep up the good work!

PR #385, opened 2023-05-20
Week 15 - Happy thoughts api

Amanda-Sjstrm:


smirrebinx (commented):
Good job! The code seems to contain the logic to work but is missing some imports and declarations (see comments).

PR #387, opened 2023-05-20
happy-thoughts-API

cecavi:


server.js
```diff
@@ -16,12 +17,83 @@ const app = express();
...
 // Start defining your routes here
+// GET: List of endpoints
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.json(listEndpoints(app))
 });
 
+// GET: List of 20 most recent thoughts, in descending order.
+app.get("/thoughts", async (req, res) => {
+  const thoughts = await Thought.find().sort({ createdAt: 'desc' }).limit(20).exec();
```
majazimnoch, line 50: Is desc working for you? I found an explanation that  In JavaScript, desc is not a valid keyword for sorting in MongoDB. Instead, I used -1 to sort in descending order based on the createdAt field. But looks like it works for you!

majazimnoch (commented):
the code is easy to follow and it looks like you did a great job, also huge congrats on connecting with front end.

PR #392, opened 2023-05-21
Happy Thoughts API

SofiaGerdmar:


server.js
```diff
@@ -15,12 +15,97 @@ const app = express();
...
+    default: Date.now
+  }
+})
+
+const HappyThought = mongoose.model("HappyThought", HappyThoughtSchema);
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(listEndpoints(app));
```
mvfrid, line 45: Nice job on adding the listEndpoints to the first page of the API! It looks great.

server.js
```diff
@@ -15,12 +15,97 @@ const app = express();
...
+const listEndpoints = require('express-list-endpoints')
+
+const { Schema } = mongoose;
+const HappyThoughtSchema = new Schema({
+  message: {
+    // type is the most important one
+    type: String,
+    required: true,
+    minlength: 5,
+    maxlength: 140,
```
mvfrid, line 27: Excellent to implement validation in the backend as well!

server.js
```diff
@@ -15,12 +15,97 @@ const app = express();
...
+const HappyThought = mongoose.model("HappyThought", HappyThoughtSchema);
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(listEndpoints(app));
 });
 
+app.get("/thoughts", async (req, res) => {
+    const thoughts = await HappyThought.find().sort({createdAt: 'desc'}).limit(20).exec();
```
mvfrid, line 49: Good job on sorting and limiting the amount of messages!

mvfrid (commented):
This looks so good Sofia! Cute design, and everything is responsive and validated in a good way. 

Nice job on the backend side also, the code is easy to follow and does what it should. 

If you want to play around a bit with the backend I would suggest adding pagination and trying to tie it into the frontend. I think you could make it look great on the frontend side as well :)

Gold star to you!

mvfrid (commented):
This looks so good Sofia! Cute design, and everything is responsive and validated in a good way. 

Nice job on the backend side also, the code is easy to follow and does what it should. 

If you want to play around a bit with the backend I would suggest adding pagination and trying to tie it into the frontend. I think you could make it look great on the frontend side as well :)

Gold star to you!

PR #401, opened 2023-05-21
Peki's api

akipeki:


server.js
```diff
@@ -1,27 +1,184 @@
...
+  }
+});
+
+// Defining the "/thoughts/:_id" GET route. It responds with a single thought based on the id
+app.get('/thoughts/:_id', async (req, res) => {
+  try {
+    const singleThought = await Thought.findById(req.params._id);
+    if (singleThought) {
+      res.status(200).json({
+        message: "One single thought",
```
AnnikaSonnek, line 97: Nice! Adding this message "One single thought" is a good way of being clear to the user about what kind of result they're getting!

server.js
```diff
@@ -1,27 +1,184 @@
...
+        success: true,
+        body: singleThought
+      });
+    } else {
+      res.status(404).json({
+        success: false,
+        body: {
+          message: "Couldn't find one single thought. Double check the ID you are using"
```
AnnikaSonnek, line 105: This is also a good message for the user! So that they know what they're doing wrong.

AnnikaSonnek (commented):
This code was easy to follow. In the beginning there were a lot of comments, which was great! Towards the end there wasn't as many comments, on one hand the code has the same structure as the code in the beginning so it's not needed but for the new functions I might have been nice to have some comments. Great job!!

PR #402, opened 2023-05-21
Week 15 project

NinaWald:


EvaSjoholm (commented):
Really well done Nina, I even saw you using map and limit function in your backend which for me is impressive! Checked the front-end in the workings and all is up to date so, good job!

PR #403, opened 2023-05-21
Week 15: Alexander Gabor

Alexander-Gabor:


TessAquilon (commented):
Hi there! 
Well done for completing Week 15, Happy Thoughts API! 🌟 The code looks very similar to mine, and there really isn't much to say here, except that you might want to clean up the commented out parts taken from what I understood were the live sessions and code-alongs. Another thing is that I couldn't find a link to your deployed API, but since you showed it during the demo on Monday I know it works. So maybe you could update the READ:ME of this project for an extra gold star 🌟This was probably the most boring code review you've read ever, so I apologize for my lack of true creativity here. This is all I got :) Keep up the good work Alexander 🙌

PR #404, opened 2023-05-21
Happy Thoughts API , Maja 

majazimnoch:


SandraMadeleine (commented):
Your code looks really nice Maja. It has a good structure and contains the requested endpoints and errors. Good job!

PR #405, opened 2023-05-21
Week 15 Happy thoughts API

SandraMadeleine:
https://project-happy-thoughts-api-qewtniwmfq-lz.a.run.app/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,8 @@
         "@babel/preset-env": "^7.16.11",
         "cors": "^2.8.5",
         "express": "^4.17.3",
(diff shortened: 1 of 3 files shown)


EmmaHoltegaard (commented):
Good job on this project! At first, I couldn't get the GET request on /thoughts to work (I got your error message "Thoughts not found"), but I think that was just because your database might have been empty? It seems like we kind of start over with the database, once we use our connection-string from Atlas. Once I posted a thought with postman it showed up just fine. 
So everything works, and you've definitely met the minimum requirements for the project :)

PR #406, opened 2023-05-22
Oscar's project happy thoughts API

OscarSindemark:
https://project-happy-thoughts-api-t7sozngaeq-uc.a.run.app

### DIFF
--- a/server.js
+++ b/server.js
@@ -34,7 +34,7 @@ const ThoughtSchema = new Schema ({
     //removes unnecessary white spaces from string
     trim: true
   },
-  hearts: {
+  likes: {
     type: Number,
     default: 0
   },
--- a/server.js
+++ b/server.js
@@ -2,7 +2,7 @@ import express from "express";
 import cors from "cors";
 import mongoose from "mongoose";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-happy-thoughts-api";
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -21,6 +21,71 @@ app.get("/", (req, res) => {
...
+  }
+})
+
+app.post('/thoughts/:thoughtId/like', async (req, res) => {
+  const { thoughtId } = req.params;
+  try {
+   const updateLikes = await Thought.findByIdAndUpdate(thoughtId, { $inc: { likes: 1 }});
+   res.status(200).json({
+    success: true, 
+    response: `Thought: ${updateLikes.message} has updated likes`
```
AntoniaGranit, line 79: I really like this response message, it's a nice touch

server.js
```diff
@@ -21,6 +21,71 @@ app.get("/", (req, res) => {
...
+      response: e,
+      message: 'Could not save thought to database'
+    });
+  }
+})
+
+app.post('/thoughts/:thoughtId/like', async (req, res) => {
+  const { thoughtId } = req.params;
+  try {
+   const updateLikes = await Thought.findByIdAndUpdate(thoughtId, { $inc: { likes: 1 }});
```
AntoniaGranit, line 76: in this snippet you increase "likes" but in the thoughts schema the likes are represented by "hearts", which is probably why this endpoint doesn't work. this is my code for this endpoint:

      const singleThought = await Thought.findById(req.params._id);
      if (singleThought) {
        singleThought.hearts += 1;
        await singleThought.save();

AntoniaGranit (approved):
Hi Oscar, sorry for the late code review lol! I tested your API in Postman and everything worked but the like endpoint, probably because either it should be named "hearts" in the endpoint code OR you should just change the name in the schema from "hearts" to "likes".

PR #408, opened 2023-05-22
Happy thoughts API - Vio

code-and-cats:
https://project-happy-thoughts-api-m6dxuape5q-lz.a.run.app/thoughts

Frontend: https://enchanting-halva-88064a.netlify.app/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,16 @@
 # Project Happy Thoughts API
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This week's project was to use Express and MongoDB to build an API which includes both GET request endpoints to return data and POST endpoints to create data. It is now also connected to the week 7 project "Happy Thoughts".
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+We got exact instructions on endpoints:
+GET /thoughts - This endpoint should return a maximum of 20 thoughts, sorted by createdAt to show the most recent thoughts first.
+POST /thoughts - This endpoint should be used to add a new post.
+POST thoughts/:thoughtId/like -  Given a valid id in the URL, the API should find that post, and update its hearts property to add one heart.
 
 ## View it live
 
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,13 +1,16 @@
...
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+This week's project was to use Express and MongoDB to build an API which includes both GET request endpoints to return data and POST endpoints to create data. It is now also connected to the week 7 project "Happy Thoughts".
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+We got exact instructions on endpoints:
```
JosefinRobertsson, line 7: Nice to include a detail of the endpoint requirements in your ReadMe, helps to explain the choices made in the code.

server.js
```diff
@@ -16,11 +16,90 @@ const app = express();
...
+    type: Date,
+    default: Date.now
+  }
+})
+
+const Post = mongoose.model("Post", PostSchema)
+
 // Start defining your routes here
 app.get("/", (req, res) => {
   res.send("Hello Technigo!");
```
JosefinRobertsson, line 42: I would possibly have changed this message to something more related to the project, in case a user goes to this endpoint

server.js
```diff
@@ -16,11 +16,90 @@ const app = express();
...
+// This endpoint should return a maximum of 20 thoughts, sorted by createdAt to show the most recent thoughts first.
+app.get("/thoughts", async (req, res) => {
+  try {
+    const post = await Post.find().sort({createdAt: 'desc'}).limit(20).exec()
+    res.status(200).json({
+      success: true,
+      response: post,
+      message: "Thoughts retrieved!"
+    })
+  } catch (error) {
```
JosefinRobertsson, line 54: I like that you type out 'error' instead of just an 'e', it doesn't hurt to be clear about what is referred to

JosefinRobertsson (commented):
The code is fulfilling the requirements, it is not possible to add likes when a thought is created, a timestamp is added both in backend and frontend and also modified to be easily readable to the user, all required endpoints are there. 
 I would not have minded some more user feedback when it comes to interacting with your api; a pointer cursor when hovering a clickable area for instance would communicate to the user that something can be done here.
Well done finding and making the necessary changes to hook your new backend up with the older frontend. Everything is working, your code is easy to read, another project well done! Bonus point for matching favicon

PR #431, opened 2023-12-01
Happy thoughts API - Inna Kokic

InnaKuljanac:
View the API live here: https://happy-thoughts-api-atw3.onrender.com
View the live site here: https://happy-thoughts-project-by-inna.netlify.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,9 +2,31 @@ import express from "express";
 import cors from "cors";
 import mongoose from "mongoose";
 
+
 const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
 mongoose.connect(mongoUrl, { useNewUrlParser: true, useUnifiedTopology: true });
 mongoose.Promise = Promise;
+
+const Thought = mongoose.model('Thought', {
+  message: {
+    type: String,
+    required: true, 
+    minlength: 5,
+    maxlength: 140,
+  },
+  hearts: {
+    type: Number,
+    default: 0,
+  },
+  createdAt: {
+    type: Date,
+    default: () => new Date()
+  }
+})
+
+// createdAt - the time the Thought was added to the database
+// Defaults to the current time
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -2,23 +2,84 @@ import express from "express";
...
 app.use(express.json());
 
 // Start defining your routes here
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  res.send(listEndpoints(app));
+});
+
+app.get("/thoughts", async (req, res) => {
+  const thoughts = await Thought.find().sort({createdAt: 'desc'}).limit(20).exec();
```
AnnaRobertsson, line 48: Good to take care of limiting it to the 20 most recent posts.

server.js
```diff
@@ -2,23 +2,84 @@ import express from "express";
...
+
+// New route to handle increasing hearts for a specific thought
+app.post("/thoughts/:thoughtId/like", async (req, res) => {
+  const { thoughtId } = req.params;
+
+  try {
+    const updatedThought = await Thought.findByIdAndUpdate(
+      thoughtId,
+      { $inc: { hearts: 1 } }, // Increment the hearts field by 1
+      { new: true } // Return the updated document
```
AnnaRobertsson, line 72: It's interesting to see new ways to do the same thing that I also did! Here I used
`updateLike.hearts += 1; // Increment the 'hearts' property of the thought
    await updateLike.save(); // Save the updated thought`
Both works great but looks very different from eachother! Fun thing about code 😄

AnnaRobertsson (commented):
I tried out your Happy Thoughts page and it worked very well, both posting and liking posts. Also just looking at your endpoints in the backend works well too. Since it takes a while for the posts to load, maybe a nice feature could be a loader/text that indicates something is loading and will appear shortly. Other than that it looks great!

The code looks tidy. Well done! 🥳

PR #437, opened 2023-12-02
Happy thought API - Frida Kristiansson

FridaKristian:
https://love-twitter-api.onrender.com/

### DIFF
--- a/server.js
+++ b/server.js
@@ -15,6 +15,10 @@ const ThoughtSchema = new Schema({
     minlength: 5,
     maxlength: 140
   },
+  username: {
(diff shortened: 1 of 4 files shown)


server.js
EmmyLJ, line 1: Your code is easy to follow and is very well structured. Some parts we've made pretty much the same. :) Cool that you've done a part that makes it possible for the user to choose if they'd like to be anonymous or use a username. When I checked out the deployed API page it worked fine. Great job! It really shows that you know what you´re doing! :D

ghost (approved):

PR #438, opened 2023-12-02
Klaudia W

klaudiaW0410:
https://project-happy-thoughts-api-vl6m.onrender.com


andrkavr (approved):
Solid, well structured code. 
Everything looks good and my only comment is that you could try using separate files for routes and models in upcoming projects.
Great job!

PR #443, opened 2023-12-03
Wk15 - Project Happy Thoughts - Beckie Morton

BeckieMorton:


mikaeloling (approved):
Nice work Beckie! I like how you have structured you endpoints and also made sure to implement good error handling. Also, the frontend looks and works nicely. I noticed that these onrender deploys usually are slow on the first load, so it might me a good practice to start implementing load screens by default (this certainly applies to my projects).

PR #449, opened 2023-12-03
Week 15 - Happy Thoughts API  - andrkavr

andrkavr:
Did this week's project as a mono-repo so the original Happy Thoughts project will still be linked to the Technigo API. 
Links to both front-end and back-end are in the readme.


backend/models/User.js
```diff
@@ -0,0 +1,36 @@
...
+    // posts: {
+    //     type: String
+    // }
+  },
+  {
+    timestamps: true,
+  }
+);
+
+export const UserModel = mongoose.model("users", userSchema);
```
majoh23, line 36: Some work has been done on user models etc, but it’s not used?

majoh23 (commented):
Good job! Takes forever to load. I guess it’s the free tier of Render that is the problem?

PR #470, opened 2024-05-16
Happy Thoughts API - Linda

linda-f:
## Netlify link
https://linda-f-project-happy-thoughts-api.onrender.com

## Collaborators
[linda-f]


server.js
```diff
@@ -16,9 +25,104 @@ const app = express();
...
+      .status(500)
+      .send(
+        "Sorry, this page is not available at the moment. Please try again later."
+      );
+  }
+});
+
+//POST the thoughts endpoint
+app.post("/thoughts", async (req, res) => {
+  const { message, hearts, createdAt, _id } = req.body;
```
wwenzz, line 82: do we need to provide "_id" for posting a thought? I don't think we need to extract that from request body as MongoDB will automatically assign that to new data.

server.js
```diff
@@ -16,9 +25,104 @@ const app = express();
...
 app.use(express.json());
 
-// Start defining your routes here
+// API documentation
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  try {
+    const endpoints = expressListEndpoints(app).map((endpoint) => {
+      if (endpoint.path === "/thoughts") {
+        endpoint.query = {
```
wwenzz, line 33: nice customised descriptions!

server.js
```diff
@@ -16,9 +25,104 @@ const app = express();
...
+  const { message, hearts, createdAt, _id } = req.body;
+
+  try {
+    const thought = await new Thought({
+      message,
+      hearts,
+      createdAt,
+      _id,
+    }).save();
+    res.status(201).json(thought);
```
wwenzz, line 91: Screenshot 2024-05-20 at 12 21 39
Great that you tried to use some mongoose method in the Schema to prevent data from being manipulated. But I think "setDefaultsOnInsert" method only works when you update the data by using "update() or findOneAndUpdate()". I also tested it out and noticed it will not assign the hearts to default 0 when I provided the hearts number (see the screenshot).
An easiest way to fix this is to extract the info you want, in this case only the message, and ignore those you don't want.

server.js
```diff
@@ -16,9 +25,104 @@ const app = express();
...
+//GET the thoughts endpoint
+//Also sorting them in a decending order and limit the thoughts to 20 shown
+app.get("/thoughts", async (req, res) => {
+  try {
+    const allThoughts = await Thought.find()
+      .sort({ createdAt: -1 })
+      .limit(20)
+      .exec();
+
+    if (allThoughts.length > 0) {
```
wwenzz, line 65: great and detailed error handling!

server.js
```diff
@@ -16,9 +25,104 @@ const app = express();
...
+    const likeThought = await Thought.findByIdAndUpdate(
+      thoughtId,
+      { $inc: { hearts: 1 } },
+      { new: true, runValidators: true }
+    );
+
+    if (likeThought) {
+      res.status(200).json(likeThought);
+    } else {
+      res
```
wwenzz, line 115: I understand that you want to customise error handling when id is not found, which is great, but the fact is the 404 line of code will never run because if no post is found with a given id, the findByIdAndUpdate will immediately throw an error instead of returning null/undefined falsy value. Then when you input invalid id it will always show 500 error you defined in the catch block.

wwenzz (commented):
Well done Linda! You implemented several error handling and also customised them, which is great 👍 The code is neat and easy to understand, I do enjoy reviewing your code!
Only the post part might need to fix, otherwise it's all good!

HIPPIEKICK (approved):
Good job! Consider formatting the times in the frontend to a more user friendly format 👇 

Skärmavbild 2024-05-22 kl  15 00 13

PR #476, opened 2024-05-17
Martin - Happy Thoughts API

Martin-Joensson:
## Netlify link
Back end: https://happy-thoughts-api-uu1o.onrender.com/
Front end: https://manyhappythoughts.netlify.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,8 +1,9 @@
 import cors from "cors";
 import express from "express";
 import mongoose from "mongoose";
+import expressListEndpoints from "express-list-endpoints";
 
-const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/project-mongo";
+const mongoUrl = process.env.MONGO_URL || "mongodb://localhost/HappyThoughts";
 mongoose.connect(mongoUrl);
 mongoose.Promise = Promise;
 
@@ -16,19 +17,48 @@ const app = express();
 app.use(cors());
 app.use(express.json());
 
-// Start defining your routes here
-app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+// Mongoose model
+const Thought = mongoose.model("Thought", {
+  message: {
+    type: String,
+    required: true,
+    maxlength: 140,
+    minlength: 5,
+  },
+  hearts: { type: Number, default: 0 },
+  createdAt: { type: Date, default: Date.now },
 });
 
-app.get("/thoughts", (req, res) => {
-  // Get all the thoughts
-})
+// Start defining your routes here
+app.get("/", (req, res) => {
+  const endpoints = expressListEndpoints(app);
+  res.json(endpoints);
+});
 
-app.post("/thoughts", (req, res) => {
-  // Create thought in the database
-  
-})
+app.get("/thoughts", async (req, res) => {
+  // Get all the thoughts
+  const thoughts = await Thought.find({}).sort(
+    { createdAt: "asc" }.limit(20).exec()
+  );
+
+  try {
+    res.status(201).json(thoughts);
+  } catch (err) {
+    res.status(400).json({
+      message: "Could not retrieve the Happy Thoughts.",
+      error: err.errors,
+    });
+  }
+});
+
+app.post("/thoughts/", async (req, res) => {
+  // Create a thought in the database
+});
+
+app.post("/thoughts/:thoughtId/like", async (req, res) => {
+  // Create like in the database
+
+});
 
 // Start the server
 app.listen(port, () => {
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -16,9 +20,66 @@ const app = express();
...
 app.get("/", (req, res) => {
-  res.send("Hello Technigo!");
+  const endpoints = expressListEndpoints(app);
+  res.json(endpoints);
+});
+
+app.get("/thoughts", async (req, res) => {
+  // Get all the thoughts
+  const thoughts = await Thought.find().sort(
+    { createdAt: "desc" }).limit(20).exec();
```
almaherris, line 44: I might be completely wrong here, but I just reacted because I did it differently. If there are errors thrown by "await Thought.find()", would they be handled? I'm thinking since it's outside of the try/catch?

server.js
```diff
@@ -16,9 +20,66 @@ const app = express();
...
+
+  try {
+    res.status(201).json(thoughts);
+  } catch (err) {
+    res.status(400).json({
+      message: "Could not retrieve the Happy Thoughts.",
```
almaherris, line 50: Interesting, we used different status codes here 😄

In "try" I used 200 as I read it's a more general "OK" status for GET requests while 201 specifically means "created" when you have done a successful POST request. 

In "catch" I used 500 which indicates server-side errors (such as database query error, ,which would likely be the case here). 

But again, might be wrong here too! By no means I feel comfortable with these status codes, just find it interesting that we chose different ones :)

almaherris (commented):
Hej Martin! Your API is working as it should and your code looks clean! We used different status codes for the GET request so I just pointed that out. Good job! 😃 💌

server.js
```diff
@@ -16,9 +20,66 @@ const app = express();
 app.use(cors());
 app.use(express.json());
 
+// Mongoose model
+const Thought = mongoose.model("Thought", {
+  message: {
+    type: String,
+    required: true,
+    maxlength: 140,
+    minlength: 5,
```
HIPPIEKICK, line 29: Why maxlength before minlength? My eeeeyes 😅

server.js
```diff
@@ -16,9 +20,66 @@ const app = express();
...
+  },
+  hearts: { type: Number, default: 0 },
+  createdAt: { type: Date, default: Date.now },
```
HIPPIEKICK, line 32: Format these lines as you did with the message for better readability

server.js
HIPPIEKICK: I'd also put it in the try block 👍

server.js
```diff
@@ -16,9 +20,66 @@ const app = express();
...
+app.post("/thoughts/", async (req, res) => {
+  // Create a thought in the database
+  const { message, heart } = req.body;
+  const thought = new Thought({ message, heart });
+  try {
+    const newThought = await thought.save();
+    res.status(201).json(newThought);
+  } catch (err) {
+    res
+      .status(500)
```
HIPPIEKICK, line 65: Can you think of another error code than 500 here? 👀

server.js
```diff
@@ -16,9 +20,66 @@ const app = express();
...
+  }
+});
+
+app.post("/thoughts/", async (req, res) => {
+  // Create a thought in the database
+  const { message, heart } = req.body;
+  const thought = new Thought({ message, heart });
+  try {
+    const newThought = await thought.save();
+    res.status(201).json(newThought);
```
HIPPIEKICK, line 62: 👍

server.js
```diff
@@ -16,9 +20,66 @@ const app = express();
...
+    res.status(201).json(thoughts);
+  } catch (err) {
+    res.status(400).json({
+      message: "Could not retrieve the Happy Thoughts.",
+      error: err.errors,
+    });
+  }
+});
+
+app.post("/thoughts/", async (req, res) => {
```
HIPPIEKICK, line 56: Minor thing, but no need for the second `/`

server.js
```diff
@@ -16,9 +20,66 @@ const app = express();
...
+      .json({ message: "Could not save your thought.", error: err.errors });
+  }
+});
+
+app.post("/thoughts/:thoughtId/like", async (req, res) => {
+  // Create like in the database
+  const { thoughtId } = req.params;
+
+  try {
+    const thought = await Thought.findByIdAndUpdate(thoughtId, {$inc: {hearts: 1}}, {new: true, runValidators: true});
```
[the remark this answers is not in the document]
HIPPIEKICK: Agreed! Just missing a bit of formatting 😇

HIPPIEKICK (approved):
Good job Martin! I just added some small comments to think about going forward 🌻

PR #477, opened 2024-05-17
project-happy-thoughts-api

lunek1:
## Render & Netlify link
Backend: https://project-happy-thoughts-api-qgyf.onrender.com/
Frontend: https://celebrated-shortbread-0673ce.netlify.app/

## Collaborators
Solo


server.js
```diff
@@ -1,24 +1,105 @@
...
-// Add middlewares to enable cors and json body parsing
 app.use(cors());
 app.use(express.json());
 
-// Start defining your routes here
+// GET /thoughts endpoint to return 20 results ordered by createdAt in descending order
+app.get("/thoughts", async (req, res) => {
+  try {
+    const thoughts = await Thought.find().sort({ createdAt: -1 }).limit(20);
+    res.json(thoughts);
```
HIPPIEKICK, line 38: Would be nice with a status code here too

server.js
```diff
@@ -1,24 +1,105 @@
...
+  }
+});
+
+// GET /thoughts/oldest endpoint to return 20 oldest thoughts ordered by createdAt in ascending order
+app.get("/thoughts/oldest", async (req, res) => {
+  try {
+    const oldestThoughts = await Thought.find()
+      .sort({ createdAt: 1 })
+      .limit(20);
+    res.json(oldestThoughts);
```
HIPPIEKICK, line 50: and here (status code)

server.js
```diff
@@ -1,24 +1,105 @@
...
+  try {
+    const thoughts = await Thought.find().sort({ createdAt: -1 }).limit(20);
+    res.json(thoughts);
+  } catch (error) {
+    res.status(500).json({ error: "Internal Server Error" });
+  }
+});
+
+// GET /thoughts/oldest endpoint to return 20 oldest thoughts ordered by createdAt in ascending order
+app.get("/thoughts/oldest", async (req, res) => {
```
HIPPIEKICK, line 45: Better to use query params for these types of queries. Then you'd only need one endpoint that's named after what it returns (/thoughts) and the user can choose whether or not they want to add a sorting query, i.e. /thoughts would work but /thoughts?sort=desc would also work

server.js
```diff
@@ -1,24 +1,105 @@
...
+    return res
+      .status(400)
+      .json({ error: "Text must be between 5 and 140 characters long" });
+  }
+  try {
+    const thought = new Thought({ text });
+    await thought.save();
+    res.status(201).json(thought);
+  } catch (error) {
+    res.status(500).json({ error: "Internal Server Error" });
```
HIPPIEKICK, line 69: Could you think of another error code to use here? 👀

server.js
```diff
@@ -1,24 +1,105 @@
...
+
+// Endpoint to add hearts to a thought
+app.post("/thoughts/:id/like", async (req, res) => {
+  const { id } = req.params;
+  try {
+    const thought = await Thought.findById(id);
+    if (!thought) {
+      return res.status(404).json({ error: "Thought not found" });
+    }
+    thought.hearts += 1;
```
HIPPIEKICK, line 81: Here you could also use $inc

server.js
```diff
@@ -1,24 +1,105 @@
...
+    await thought.save();
+    res.json(thought);
+  } catch (error) {
+    res.status(500).json({ error: "Internal Server Error" });
+  }
+});
+
+// GET /thoughts/likes endpoint to retrieve all thoughts sorted by most likes to lowest
+app.get("/thoughts/likes", async (req, res) => {
```
HIPPIEKICK, line 90: Same with this, GET endpoints should be named after what they return, so this could also be a query under the /thoughts endpoint

HIPPIEKICK (approved):
Good job! Wrote some pointers on best practices to think about going forward 🌻 

PS. In your frontend, you probably want to trigger a re-render after a new though has been posted

PR #483, opened 2024-05-18
Erika's Happy Thoughts API

Tejpex:
## Netlify link
https://tejpex-happy-thoughts-api.onrender.com/

## Collaborators
Add your collaborators here. Write their GitHub usernames in square brackets. If there's more than one, separate them with a comma, like this:
[github-username-1, github-username-2]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,15 @@
 # Project Happy Thoughts API
 
-Replace this readme with your own information about your project.
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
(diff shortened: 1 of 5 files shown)


HIPPIEKICK (approved):
Good job and nice to see that you made it your own with the dislike endpoint. What do you think about having it as one endpoint and then using a query to specify exactly what the user wants to do? Something like:
/thoughts/mongoID?like=true
and
/thoughts/mongoID?like=false

or something, maybe with better naming. Just a thought.

Good work!

PR #485, opened 2024-05-18
Wen's happy thoughts API

wwenzz:
## Netlify link
Frontend: https://wen-happy-thoughts.netlify.app/
Backend: https://wen-happy-thoughts-api.onrender.com/

## Collaborators
[wwenzz]

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,6 +2,7 @@ import cors from "cors";
 import express from "express";
 import expressListEndpoints from "express-list-endpoints";
 import mongoose from "mongoose";
+
 import Thought from "./models/Thought";
 
 const mongoUrl =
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -18,7 +22,89 @@ app.use(express.json());
...
+    res.status(400).json({
+      success: false,
+      error: error.message,
+      message: "Could not like the thought...",
+    });
+  }
+});
+
+//unlike a post
+app.post("/thoughts/:thoughtId/unlike", async (req, res) => {
```
HIPPIEKICK, line 82: Love this! What do you think about having both the like and dislike in the same endpoint though? Maybe some kind of param differing them, since both are changing the likes?

HIPPIEKICK (approved):
Nice job Wen, looks super! Nice that you created dislike functionality as well. Makes sense!

PR #490, opened 2024-05-20
project-happy-thoughts-api-ji

jyy009:
### Backend:
https://project-happy-thoughts-api-w357.onrender.com/

### Frontend:
https://polite-begonia-9a20c4.netlify.app/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ node_modules
 .env.test.local
 .env.production.local
 package-lock.json
-instructions.md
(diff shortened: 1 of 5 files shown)


HIPPIEKICK (changes requested):
Tried out your frontend, and the like function is causing an error. Did you remember to update the frontend to PATCH instead of POST? 👀

jyy009:
I have updated the patch url!

NellaMori (approved):

PR #500, opened 2024-12-19
happy-thoughts-api

gittebe:
https://project-happy-thoughts-api-production-5d1d.up.railway.app/

### DIFF
--- a/server.js
+++ b/server.js
@@ -95,13 +95,14 @@ app.get("/thoughts", async (req, res) => {
         }
 
     // Update hearts count
-    const newHearts = thought.hearts % 2 ===0 ? thought.hearts +1 : thought.hearts -1;
-    await Thought.findByIdAndUpdate(id, {hearts: newHearts}, {new: true});
+    const newHearts = thought.hearts % 2 === 0 ? thought.hearts +1 : thought.hearts -1;
+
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -1,27 +1,113 @@
...
+    } catch(error) {
+      res.status(400).json({message: "Could not save thought", errors: error.err.errors})
+    }
+    });
+
+    // Like a thought (increase of hearts)
+    app.post("/thoughts/:id/like", async (req, res) => {
+      try {
+        const {id} = req.params
+        const thought = await Thought.findById(id)
```
HIPPIEKICK, line 92: Why not use findByIdAndUpdate for updating hearts directly instead of fetching, modifying, and saving? 🤔

server.js
```diff
@@ -1,27 +1,113 @@
...
+        if(!thought) {
+          return res.status(404).json({message: "Thought was not found"})
+        }
+        if (thought.hearts %2=== 0) {
+          thought.hearts+=1
+        } else {
+          thought.hearts -=1
+        }
```
HIPPIEKICK: What is this for? 😅

HIPPIEKICK (dismissed):
Use mongoose methods when you can, otherwise - good job!

server.js
gittebe: Thank you :) I changed it

JennieDalgren (approved):

PR #506, opened 2024-12-20
Expres.js and MongoDB-backed Happy Thoughts

HeleneWestrin:
## Links
[View API »](https://project-happy-thoughts-api-kappa.vercel.app/)
[View Frontend Site »](https://mongodb-api--happy-thoughts-by-helene.netlify.app/)

### DIFF
--- a/server.js
+++ b/server.js
@@ -50,7 +50,9 @@ const port = process.env.PORT || 8080;
 const app = express();
 
 // Add middlewares to enable cors and json body parsing
-app.use(cors());
+app.use(
+  cors({ origin: "https://mongodb-api--happy-thoughts-by-helene.netlify.app" })
+);
 app.use(express.json());
 
 // API documentation
--- a/server.js
+++ b/server.js
@@ -7,8 +7,7 @@ import { configDotenv } from "dotenv";
 // Load environment variables
 configDotenv();
 
-const mongoUrl =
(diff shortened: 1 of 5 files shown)


server.js
```diff
@@ -1,24 +1,122 @@
...
+          `Message must be at least 5 characters. Your message has ${props.value.length} characters.`,
+      },
+      {
+        validator: function (value) {
+          return value.length <= 140;
+        },
+        message: (props) =>
+          `The maximum length of a message is 140 characters. Your message has ${props.value.length} characters.`,
+      },
+    ],
```
HIPPIEKICK, line 34: ⭐

server.js
```diff
@@ -1,24 +1,122 @@
...
+    if (!updatedThought) {
+      return res.status(404).json({ error: "Thought not found" });
+    }
+
+    res.status(200).json(updatedThought);
+  } catch (err) {
+    res
+      .status(500)
+      .json({ error: "Failed to update like", details: err.message });
+  }
```
HIPPIEKICK, line 119: Good idea! ⭐ However, I think it would make sense with a PATCH req instead of PUT (since we're not replacing the whole resource)

HIPPIEKICK (approved):

server.js
HeleneWestrin: Great advice, thank you Matilda! 🙏

